PR #18, opened 2021-09-09
fix(prefixStorage): handle missing key

farnabaz:
`key` argument is optional in some functions, like (`getKeys`)

- Add default value for key argument to prevent undefined key (`base:undefined`)
- Normalize `base + key` to remove ending `:` in case of missing key (`base:`)

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -1,5 +1,5 @@
 import type { Storage } from './types'
-import { normalizeBase, normalizeKey } from './_utils'
+import { normalizeBase } from './_utils'
 
 const storageKeyProps: (keyof Storage)[] = [
   'hasItem',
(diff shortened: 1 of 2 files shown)


src/utils.ts
```diff
@@ -23,7 +23,7 @@ export function prefixStorage (storage: Storage, base: string): Storage {
   const nsStorage: Storage = { ...storage }
   for (const prop of storageKeyProps) {
     // @ts-ignore Better types?
-    nsStorage[prop] = (key: string, ...args) => storage[prop](base + key, ...args)
+    nsStorage[prop] = (key: string = '', ...args) => storage[prop](normalizeKey(base + key), ...args)
```
pi0, line 26: `normalizeBase` ensures we have `:` in the end of the base and storage methods always normalize the final key. Do you have an example of why this doesn't works? (adding to fixture would be bonous!)

farnabaz: > storage methods always normalize the final key.

You are right, `normalizeKey` is redundant here

About the default value for key:
When we call `storage.getKeys()`, the `key` will be `undefined` and therefor `base + key` will become `BASE:undefined` which is wrong behavior

PR #26, opened 2021-09-28
feat: allow driver getKey to receive base key

farnabaz:
This PR allows drivers to filter out keys in `geyKeys` method.

If our driver was mounted on `/tmp/foo` and someone calls `getKeys('/tmp/foo/bar')`, the getkeys function of driver will receive `/bar` as its argument and it can filter out the result.

This will help up to improve speed/performance for some drivers like fsDriver


pi0 (approved):

pi0:
Released as 0.2.9. Sorry for the delay

PR #34, opened 2021-11-09
fix(prefixStorage): strip keys

### LINKED ISSUE #33: `prefixStorage` should remove prefix key from `getKeys` results
### Assume
```js
/*
/tmp
/tmp/1.js
/tmp/directory/2.js
/tmp/directory/3.js
*/
  driver: fsDriver({ base: './tmp' })
})

const ps = prefixStorage(storage, 'tmp')
```

### What happening
`prefixStorage` (`ps`) keeps `tmp` as prefix in keys and so results of `getKeys` are invalid keys for `ps`
```js
const psKeys = await ps.getKeys() // return [/tmp/directory/2.js, /tmp/directory/3.js]

// but these keys cannot be read form `ps` 

await ps.getItem('/tmp/directory/2.js') // return null
```

### What Expected
`prefixStorage` (`ps`) should remove `tmp` from keys
```js
const psKeys = await ps.getKeys() // return [/directory/2.js, /directory/3.js]

// and these keys can be read form `ps` 

await ps.getItem('/directory/2.js') // return "FILE CONTENT"
```

### Suggestion
```js
export function prefixStorage (storage: Storage, base: string): Storage {
  base = normalizeBase(base)
  if (!base) {
    return storage
  }
  const nsStorage: Storage = { ...storage }
  for (const prop of storageKeyProps) {
    // @ts-ignore Better types?
    nsStorage[prop] = (key: string = '', ...args) => storage[prop](base + key, ...args)
  }
  // Suggested logic for `getKeys`
  const prefixRegex = new RegExp(`^${base}:`)
  nsStorage.getKeys = (key: string = '', ...args) =>
    storage
      .getKeys(base + key, ...args)
      // Remove Prefix
      .then(keys => keys.map(key => key.replace(prefixRegex, '')))

  return nsStorage
}
```

WDYT @pi0 ?

farnabaz:
close #33

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -25,5 +25,12 @@ export function prefixStorage (storage: Storage, base: string): Storage {
     // @ts-ignore Better types?
     nsStorage[prop] = (key: string = '', ...args) => storage[prop](base + key, ...args)
   }
+  const prefixRegex = new RegExp(`^${base}`)
+  nsStorage.getKeys = (key: string = '', ...args) =>
+    storage
+      .getKeys(base + key, ...args)
+      // Remove Prefix
+      .then(keys => keys.map(key => key.replace(prefixRegex, '')))
+
   return nsStorage
 }
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -25,12 +25,11 @@ export function prefixStorage (storage: Storage, base: string): Storage {
     // @ts-ignore Better types?
     nsStorage[prop] = (key: string = '', ...args) => storage[prop](base + key, ...args)
   }
-  const prefixRegex = new RegExp(`^${base}`)
   nsStorage.getKeys = (key: string = '', ...args) =>
     storage
       .getKeys(base + key, ...args)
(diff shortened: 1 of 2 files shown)


src/utils.ts
```diff
@@ -25,5 +25,12 @@ export function prefixStorage (storage: Storage, base: string): Storage {
     // @ts-ignore Better types?
     nsStorage[prop] = (key: string = '', ...args) => storage[prop](base + key, ...args)
   }
+  const prefixRegex = new RegExp(`^${base}`)
+  nsStorage.getKeys = (key: string = '', ...args) =>
+    storage
+      .getKeys(base + key, ...args)
+      // Remove Prefix
+      .then(keys => keys.map(key => key.replace(prefixRegex, '')))
```
pi0, line 33: Is there a case that `getKeys` returns without our expected prefix? If not we can simplify this to simply substr by length

pi0 (approved):
Sorry i missed this

PR #36, opened 2022-02-28
feat: git driver

farnabaz:
Git driver using `isomorphic-git`


Atinux:
Do you plan to support `onAuth` for private repos?

https://isomorphic-git.org/docs/en/onAuth

src/drivers/git.ts
```diff
@@ -0,0 +1,84 @@
...
+  const clone = async () => git.clone(gitOpts)
+  
+  const resolve = (...path: string[]) => join(gitOpts.dir, ...path)
+  
+  const r = (key: string) => resolve(opts.path || '/', key.replace(/:/g, '/'))
+
+  return {
+    
+    async hasItem (key) {
+      if (!initiated) {
```
pi0, line 35: Refactor: We can create a utility like `ensureInitialized()` and call in several utils. Also caching one promise to avoid cloning multiple times in a race condition

src/drivers/git.ts
```diff
@@ -0,0 +1,84 @@
+import fs, { existsSync, promises as fsp } from 'fs'
+import git from 'isomorphic-git'
+import http from 'isomorphic-git/http/node'
```
pi0, line 3: We could create an additional driver like `git.node` for node-imports. By default I would rather to use isomorphic (web) implementation

Tahul:
Hello 🙂

Any news on this PR @pi0?

We are willing to try this driver for multi-source support on @nuxt/content v2.

Thanks a lot!

pi0:
Hi @farnabaz. Coming back to this PR, I've rebased it and added a true isomorphic HTTP client but seems it also needs a fs polyfill in order to work without Node.js.

At this point, i think it doesn't worth to spend time on this driver. We might later think about supporting giget driver for lightweight cloning but that also depends if we can have a virtual fs for browser.

PR #55, opened 2022-05-18
feat: `cloudflare-kv-http` driver

### LINKED ISSUE #31: Using Cloudflare KV outside of Workers using rest API
Without binding, we can use REST API to use KV namespace.

Example impl: https://github.com/Zertz/cloudflare-kv/blob/main/index.mjs

cpsoinos:
Add `cloudflare-kv-http` driver for Cloudflare KV using Cloudflare's REST API. Allows KV to be used outside of a CF worker environment.

Uses `msw` (https://mswjs.io/) to provide a mock server for tests.

closes #31


pi0 (approved):
Beautiful work! Thanks!

PR #61, opened 2022-06-09
feat: `github` driver

farnabaz:
Create Github driver using Github API and raw CDN

- Use API to fetch files
- Use Github raw CDN to fetch file content
- Cache files to 10 minutes (configurable by `ttl`)
- Accept `dir` to load contents from specific directory

### DIFF
--- a/src/drivers/github.ts
+++ b/src/drivers/github.ts
@@ -1,0 +1,113 @@
+import { defineDriver } from './utils'
+import { $fetch } from 'ohmyfetch'
+import { withTrailingSlash } from 'ufo'
+
+export interface GithubOptions {
+  org: string
+  repo: string
+  /**
+   * @default "main"
+   */
+  branch: string
+  /**
+   * @default ""
+   */
+  dir: string
+  token?: string
+
+  /**
+   * @default 600
+   */
+  ttl: number
+}
+
+export default defineDriver((options: GithubOptions) => {
+  const { org, repo, branch = 'main', dir, token, ttl = 600 } = options
+  const rawUrl = `https://raw.githubusercontent.com/${org}/${repo}/${branch}/${dir}`
+  
+  let files = {}
+  let lastCheck = 0
+  let isInitialized = false
+  let initialTask: Promise<any>
+
+  const syncFiles = async () => {
+    if (!isInitialized || (lastCheck + ttl * 1000) < Date.now()) {
+      if (!initialTask) initialTask = fetchFiles(options)
+      files = await initialTask
+      isInitialized = true
+      lastCheck = Date.now()
+    }
+  }
+
+  return {
+    async getKeys () {
+      await syncFiles()
+
+      return Object.keys(files)
+    },
+    async hasItem (key) {
+      await syncFiles()
+
+      return !!files[key]
+    },
+    async getItem (key) {
+      await syncFiles()
+
+      const item = files[key]
+
+      if (!item) {
+        return null
+      }
+
+      if (!item.body) {
+        try {
+          item.body = await $fetch(`${rawUrl}/${key.replace(/:/g, '/')}`, {
+            headers: {
+              Authorization: token ? `token ${token}` : undefined
+            }
+          })
+        } catch (e) {
+          // eslint-disable-next-line no-console
+          console.log(`Failed to fetch file "${key}"`, e)
+        }
+      }
+      return item.body
+    },
+    async getMeta (key) {
+      await syncFiles()
+
+      const item = files[key]
+      if (!item) {
+        return null
+      }
+      return item.meta
+    }
+  }
+})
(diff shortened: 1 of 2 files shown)


src/drivers/github.ts
```diff
@@ -0,0 +1,113 @@
...
+      return item.meta
+    }
+  }
+})
+async function fetchFiles(options: GithubOptions) {
+  const { dir, token, branch, org, repo } = options
+  const prefix = withTrailingSlash(dir)
+  const files = {}
+  try {
+    const tree = await $fetch(`https://api.github.com/repos/${org}/${repo}/git/trees/${branch}?recursive=1`, {
```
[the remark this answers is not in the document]
pi0: Later improvement: We can expose storage instance to the driver and support cachePath so that github driver can permanently cache this.

src/drivers/github.ts
```diff
@@ -0,0 +1,113 @@
...
+  return {
+    async getKeys () {
+      await syncFiles()
+
+      return Object.keys(files)
+    },
+    async hasItem (key) {
+      await syncFiles()
+
+      return !!files[key]
```
pi0, line 51: Can we use `key in files`?

src/drivers/github.ts
```diff
@@ -0,0 +1,113 @@
...
+  const files = {}
+  try {
+    const tree = await $fetch(`https://api.github.com/repos/${org}/${repo}/git/trees/${branch}?recursive=1`, {
+      headers: {
+        Authorization: token ? `token ${token}` : undefined
+      }
+    })
+    tree.tree.forEach((node) => {
+      if (node.type === 'blob' && node.path.startsWith(prefix)) {
+        const key = node.path.substring(prefix.length).replace(/\//g, ':')
```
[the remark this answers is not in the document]
farnabaz: This is for internal usage, for finding keys in `files`

src/drivers/github.ts
```diff
@@ -0,0 +1,118 @@
...
+  dir: string
+  /**
+   * @default 600
+   */
+  ttl: number
+  token?: string
+}
+
+const GITHUB_URL = 'https://api.github.com'
+const GITHUB_CDN_URL = 'https://raw.githubusercontent.com'
```
pi0, line 23: Can you please add optional options to override?

src/drivers/github.ts
```diff
@@ -0,0 +1,118 @@
...
+
+export default defineDriver((_opts: GithubOptions) => {
+  const opts = { branch: 'main', ttl: 600, ..._opts }
```
pi0, line 26: I think we can keep default object outside. Also, default dir is missing?

src/drivers/github.ts
```diff
@@ -0,0 +1,118 @@
...
+  const rawUrl = `${GITHUB_CDN_URL}/${opts.repo}/${opts.branch}/${opts.dir}`
```
pi0, line 27: Can we please use `ufo.joinURL` for joining URLs?

src/drivers/github.ts
```diff
@@ -0,0 +1,118 @@
...
+
+      const item = files[key]
+
+      if (!item) {
+        return null
+      }
+
+      if (!item.body) {
+        try {
+          item.body = await $fetch(`${rawUrl}/${key.replace(/:/g, '/')}`, {
```
pi0, line 69: We can use `baseURL` option of fetch

src/drivers/github.ts
```diff
@@ -0,0 +1,118 @@
...
+      headers: {
+        Authorization: opts.token ? `token ${opts.token}` : undefined
+      }
+    })
+
+    for (const node of trees.tree) {
+      if (node.type !== 'blob') {
+        continue
+      }
+      if (!node.path.startsWith(prefix)) {
```
pi0, line 103: Any reason not joining with || with previous condition?

src/drivers/github.ts
```diff
@@ -0,0 +1,118 @@
...
+        continue
+      }
+      const key = node.path.substring(prefix.length).replace(/\//g, ':')
+      files[key] = {
+        sha: node.sha,
```
pi0, line 108: Any reason not keeping sha inside `meta`?

src/drivers/github.ts
```diff
@@ -0,0 +1,118 @@
...
+        meta: {
+          size: node.size
```
pi0, line 110: Can we also keep mode in meta?

pi0:
@farnabaz Adding to the docs would be nice also mentioning driver usage and behavior. You can use `yarn doctoc` to update TOC.

src/drivers/github.ts
farnabaz: Applied, do you mind double check in this?

PR #64, opened 2022-06-22
fix(drivers): add prefix key for cloudflare kv list operation

clemcode:
## Type
Bugfix - Cloudflare KV Bindings driver

## Current behavior
When using the Cloudflare KV Bindings driver `getKeys(base?: string)` method with a `base` argument, an error is thrown in production.

## Description
The Cloudflare KV `list` method requires an Object `{ prefix: <base>}` to filter by prefix. This PR creates this object if a `base` argument is provided to the driver's `getKeys()` method.


pi0 (approved):

pi0:
Thanks!

PR #74, opened 2022-07-14
fix(github): trim leading slash on `dir` prefix paths

rbndelrio:
This PR removes leading slashes from file tree prefixes and fixes the default config's functionality.

The Github REST API's tree paths don't contain leading slashes, so setting the `dir` option to `/docs` or leaving it as default (`''`) was causing the driver to view these trees as empty.

### DIFF
--- a/src/drivers/github.ts
+++ b/src/drivers/github.ts
@@ -111,7 +111,7 @@ export default defineDriver((_opts: GithubOptions) => {
 })
 
 async function fetchFiles(opts: GithubOptions) {
-  const prefix = opts.dir ? withTrailingSlash(opts.dir) : ''
+  const prefix = withTrailingSlash(opts.dir).replace(/^\//, '')
   const files = {}
   try {
     const trees = await $fetch(`/repos/${opts.repo}/git/trees/${opts.branch}?recursive=1`, {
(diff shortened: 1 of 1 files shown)


src/drivers/github.ts
```diff
@@ -111,7 +111,7 @@ export default defineDriver((_opts: GithubOptions) => {
 })
 
 async function fetchFiles(opts: GithubOptions) {
-  const prefix = withTrailingSlash(opts.dir)
+  const prefix = withTrailingSlash(opts.dir).replace(/^\//, '')
```
pi0, line 114: Can you please use withoutLeadingSlash utility?

rbndelrio: Unfortunately `withoutLeadingSlash`'s behavior is to add a slash to empty strings so it doesn't really fit the bill here:
```js
withoutLeadingSlash('/docs') === 'docs'
withoutLeadingSlash('') === '/'
```

Alternatively we could use a ternary?
```js
opts.dir ? withoutLeadingSlash(withTrailingSlash(opts.dir)) : ''
```

rbndelrio: Hey! So I think you're mistaken? We need to remove the leading slash while keeping the trailing one. Otherwise, a config of `dir: 'docs'` would expose files like `docs_private/Secret.md`.

pi0: Sorry i was mistaken because of title.

rbndelrio: Oh yikes my bad. Thanks for merging this in!

PR #81, opened 2022-08-21
fix(cloudflare-kv-http): hasItem and getItem

qin-guan:
The current `cloudflare-kv-http` driver implementation of `getItem` does not seem to follow the Cloudflare API response documented [here](https://api.cloudflare.com/#workers-kv-namespace-read-key-value-pair), which returns the value directly.


hexorx:
👍

src/drivers/cloudflare-kv-http.ts
```diff
@@ -96,13 +96,21 @@ export default defineDriver<KVHTTPOptions>((opts) => {
...
   }
 
   const getItem = async (key: string) => {
-    const { result } = await kvFetch(`/values/${key}`)
-    return result
+    try {
+      // Cloudflare API returns with `content-type: application/octet-stream`
+      return await kvFetch(`/values/${key}`, { parseResponse: txt => txt })
+    } catch {
+      return null
```
pi0, line 112: I think we should only assume 404 is valid response and not hide any other (token or network errors)

src/drivers/cloudflare-kv-http.ts
```diff
@@ -96,13 +96,21 @@ export default defineDriver<KVHTTPOptions>((opts) => {
   const kvFetch = $fetch.create({ baseURL, headers })
 
   const hasItem = async (key: string) => {
-    const { success } = await kvFetch(`/values/${key}`)
-    return success
+    try {
+      await kvFetch(`/values/${key}`)
```
pi0, line 100: (not directly related to this PR) We might use meta-data endpoint for existance instead of getting value.

pi0:
Thanks for fixes @qin-guan and sorry took long to review. I've pushed few improvements to use metadata (it seems initial implementation was indented to use it for success)

qin-guan:
No worries @pi0, thanks for reviewing and making the changes! 🎉

PR #82, opened 2022-09-13
feat: add unwatch functions

### LINKED ISSUE #79: Allow to stop watcher without using dispose
It seems it is not possible to cancel a watcher registered with `storage.watch()` unless we use `storage.dispose()`

It would be great to have a way to cancel either one watcher callback like:
```ts
const watcher = await storage.watch(() => {})
await watcher.stop()
```

Or all active watcher, without clearing memory data or unmounting drivers
```ts
await storage.unwatch()
```

stafyniaksacha:
Add return signature to watch function (in Driver and Storage) to remove current watcher from being proceeded, if no watcher callback are present the Storage will stop watcher from Driver.

Only one watcher is allowed per mount point, but can allow multiple handler for them. Unwatch will stop watchers only if no handler remains.

We can now either stop one watcher
```ts
const watcher = await storage.watch(() => {})
await watcher.stop()
```

Or discard all with
```ts
await storage.watch(() => {})
await storage.unwatch()
```

- close #79


pi0 (approved):
Thanks for the PR 💯

PR #86, opened 2022-10-04
feat(server): use patch and post as put

### LINKED ISSUE #85: [server] Support POST and PATCH too for adding items
Usually, `PUT` and `PATCH` are used interchangeably (e.g. in Laravel). So it'd be nice to add them as request options. As there is no "edit" item, I was in belief that `POST` should work for setting an item.

I'd propose to add `POST` and `PATCH` support too.

Happy to send in a PR.

manniL:
Resolves #85


manniL:
No intent to ship as discussed in #85

PR #92, opened 2022-10-26
Memcached support

Sanicboi:
Added memcached support. Had to change some functions and types due to the fact that they are not supported by memcached. For instance, memcached does not support getKeys.


Hebilicious (approved):
LGTM

Hebilicious:
@pi0 Memcached does not supports `getKeys. I tried using cachedump but it wasn't reliable. We could manually implement a custom solution to keep track of all the keys set with `setItem` (underneath a custom key).
But that would deservers a separated PR as its not as straightforward as it sounds. 
If there's demand for it, we can definitely do it in a later. I've implemented a way to skip the `getKeys` test for one driver, but we could disable them all for memcached as most of them use `getKeys` underneath. 

The docs clearly reflects that `getKeys` doesn't work with memcached, so I think it's fine.

PR #95, opened 2022-10-28
docs: fix typo

manniL:


danielroe (commented):
I think both words are acceptable, with `shakable` being slightly more common

manniL:
Oh, didn't know that! Thanks 😋

PR #96, opened 2022-11-13
feat: add `aws-s3` driver

Suven:
Hey,

this is my starting point for supporting S3 via the SDK. I need it myself for an API that I deploy to AWS lambda.

There are still some rough edges and testing to be done on my end, but maybe it is a good starting base for feedback. 

Things I am unsure about:

1. How can this be tested? Mocking everything? Or is there maybe a test bucket by amazon I could use?
2. Listing the keys is limited to 1k keys by the API but it is possible to filter/paginate those results. Can we provide that from within unstorage or should it just be out of scope?
3. Should I try to deserialize json or just return strings? redis doesn't unwrap, does it?
4. ~~Should I catch errors when requesting meta for nonexistint keys?~~ => seems so from looking at the fs-implementation. Will add the same behaviour
5. There is no explicit `has` in S3, so I am just doing a getMeta instead


Hebilicious:
Hi @Suven are you still working on this? Is this ready to review ?

Suven:
@Hebilicious I was kind of waiting / hoping for feedback :'D In the meantime I stopped needing it for myself.

Suven:
As I did not continue this AND there is an alternative that is more versatile, I'll close this (:

PR #138, opened 2023-02-06
fix(cloudflare): pass params to kv request

danielroe:
### ❓ Type of change

- [x] 🧹 Chore
- [ ] 📖 Documentation (updates to the documentation or readme)

### 📚 Description

Currently we spread the params within `ofetch` so we never actually end up passing any to the fetcher, meaning currently this (if the number of keys is high enough to warrant pagination) just requests the same list of keys over and over again until it times out.

### 📝 Checklist

### DIFF
--- a/src/drivers/cloudflare-kv-http.ts
+++ b/src/drivers/cloudflare-kv-http.ts
@@ -128,9 +128,9 @@ export default defineDriver<KVHTTPOptions>((opts) => {
   const getKeys = async (base?: string) => {
     const keys: string[] = []
 
-    const params = new URLSearchParams()
+    const params: Record<string, string> = {}
     if (base) {
-      params.set('prefix', base)
+      params.prefix = base
     }
 
     const firstPage = await kvFetch('/keys', { params })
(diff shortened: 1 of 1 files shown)


pi0:
Can we fix params itself to be an object? It is used for L136 as well

src/drivers/cloudflare-kv-http.ts
```diff
@@ -128,27 +128,27 @@ export default defineDriver<KVHTTPOptions>((opts) => {
...
+    while (params.cursor) {
       const pageResult = await kvFetch('/keys', { params })
       pageResult.result.forEach(({ name }: { name: string }) => keys.push(name))
       const pageCursor = pageResult.result_info.cursor
       if (pageCursor) {
-        params.set('cursor', pageCursor)
+        params.cursor = pageCursor
       } else {
-        params.delete('cursor')
+        delete params.cursor
```
pi0, line 151: small perf: avoids reshaping object

```suggestion
        params.cursor = undefined

```

pi0:
Thanks

PR #140, opened 2023-02-06
feat(driver): add planetscale driver

danielroe:


src/drivers/planetscale.ts
```diff
@@ -0,0 +1,80 @@
...
+  updated_at: Date;
+}
+
+export default defineDriver((opts: PlanetscaleDriverOptions = {}) => {
+  if (!opts.url)
+    throw new Error("Database URL is required to use the Planetscale driver.");
+
+  opts.table = opts.table || "storage";
+
+  const c = connect({
```
pi0, line 24: Perf: Can we make connection lazy same as other drivers?

danielroe: It is lazy. This only configures the connection.

pi0: See see checking bundle code. Very confusing naming! What do you think to directly use `new Connection()` here for code clarity that it is not initiating a connection? 

(BTW it _might_ do sync operations in future versions of library or at least try a background connect open init. explicit lazy getter ensures this won't happen)

danielroe: I think using `new Connection()` would be just as likely to perform sync operation in future. We could add a comment if you think it will help 😆

FWIW, I don't think it is likely as the purpose of this library is specifically for use in edge functions: https://github.com/planetscale/database-js.

pi0: The wrapper is much better IMO to avoid the risks if you don't mind to add. but if prefer to keep like this, also a comment would help so at least anyone reading this driver code as example quickly understands this is not making any connection. `connect()` is basically wrapper for `new Connection`

danielroe: I have added the comment already. Feel free to add the lazy wrapper but I would personally prefer not.

src/drivers/planetscale.ts
```diff
@@ -0,0 +1,86 @@
...
+
+  // `connect` configures a connection class rather than initiating a connection
+  const c = connect({
+    url: opts.url,
+    fetch,
+  });
+
+  // this query will be lazily executed
+  if (opts.cache) {
+    c.execute("SET @@boost_cached_queries = true;");
```
pi0, line 33: Are you sure it is lazy? Both return type (c.execute returns promise) and implementation i can find in `@planetscale/database` are returning promise:

image

danielroe: Yes, but we don't await it.

pi0: It is not lazy executed only background this way... And will cause the server to make a post request on server startup/initializing drivers even if we don't need it immediately.

danielroe: What would you prefer? We can drop the feature or wrap this execution?

pi0: Can you please verify last changes?

danielroe: looks perfect to me 👌

pi0 (approved):

PR #143, opened 2023-02-06
fix: update driver defenition types

danielroe:


src/types.ts
```diff
@@ -14,10 +14,10 @@ export interface StorageMeta {
 
 export interface Driver {
   hasItem: (key: string) => MaybePromise<boolean>;
-  getItem: (key: string) => StorageValue;
+  getItem: (key: string) => MaybePromise<StorageValue>;
   /** @experimental */
-  getItemRaw?: (key: string) => unknown;
-  setItem?: (key: string, value: string) => MaybePromise<void>;
+  getItemRaw?: (key: string) => MaybePromise<unknown>;
+  setItem?: (key: string, value: StorageValue) => MaybePromise<void>;
```
pi0, line 20: Currently, we actually always stringify for setter.

```suggestion
  setItem?: (key: string, value: string) => MaybePromise<void>;
```

PR #145, opened 2023-02-06
feat(driver): add basic `cache` driver

danielroe:
### 🔗 Linked issue

maybe https://github.com/unjs/unstorage/issues/15 ?

### ❓ Type of change

- [ ] 📖 Documentation (updates to the documentation or readme)
- [ ] 🐞 Bug fix (a non-breaking change that fixes an issue)
- [x] 👌 Enhancement (improving an existing functionality like performance)
- [ ] ✨ New feature (a non-breaking change that adds functionality)
- [ ] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

### 📚 Description

This is a driver that can wrap another drive to provide automatic caching of `hasItem`, `setItem` and `getItem`

### 📝 Checklist

- [x] I have linked an issue or discussion.
- [x] I have updated the documentation accordingly.

Issue #15, opened 2021-09-08
Cached Storage


pi0:

chris-visser:
Not sure wjat you mean with this, but I had this idea to allow setting the cache via s3 endpoints and then expose it via a cloudfront distribution. 

I'm currently attempting to create the s3 driver. Wdyt of adding a cdn option to drivers like s3 and maube other bucket providers?

tanishqmanuja:
For now I am using something like this, are there any downsides to this approach ?

```ts
import { setTimeout } from "node:timers/promises";

import { createStorage } from "unstorage";
import { fsCacheDriver } from "./driver";

const storage = createStorage({
  driver: fsCacheDriver({ base: "./db", freshness: "5s" }),
});

await storage.setItem("foo", "bar");

let foo = await storage.getItem("foo");
console.log(foo); // output - bar

await setTimeout(5000);

foo = await storage.getItem("foo");
console.log(foo); // output - undefined
```
<details>

<summary>driver.ts</summary>

```ts
import fsLiteDriver, { type FSStorageOptions } from "unstorage/drivers/fs-lite";
import { normalize, type Duration } from "./duration";

export interface FSCacheStorageOptions extends FSStorageOptions {
  freshness?: Duration;
}

const DRIVER_NAME = "fs-cache";
const DEFAULT_EXPIRATION = "30m";

export const fsCacheDriver = (options: FSCacheStorageOptions = {}) => {
  const driver = fsLiteDriver(options);
  const freshness = normalize(options.freshness ?? DEFAULT_EXPIRATION);

  return {
    ...driver,
    name: DRIVER_NAME,
    options,
    async getItem(key: string, opts: { freshness?: Duration } = {}) {
      const item = await driver.getItem(key);

      if (item === null || item === undefined) {
        return item;
      }

      const meta = await driver.getMeta?.(key, {});

      if (meta === null || meta === undefined || meta.mtime === undefined) {
        return item;
      }

      const etime = opts.freshness ? normalize(opts.freshness) : freshness;

      if (Date.now() - new Date(meta.mtime).getTime() < etime) {
        return item;
      }

      if (!options.noClear) {
        await driver.removeItem?.(key, {});
      }
    },
  };
};
```

</details>

<details>

<summary>duration.ts</summary>

```ts
export const UNITS = {
  s: 1000,
  m: 60000,
  h: 3600000,
  d: 86400000,
  w: 604800000,
} as const;

export type Unit = keyof typeof UNITS;

[...]

Issue #13, opened 2021-09-08
Support native ttl


pi0:

tiborpino:
@pi0 Any progress here?

McPizzaIs:
for Redis, this could be addressed by #150

pi0:
Thinking more about the generic TTL idea, the new transaction options feature (https://github.com/unjs/unstorage/pull/168) can allow any driver to implement it per `setItem` call in their native way.


Also drivers can globally support ttl. For memory, you can use [LRU Driver](https://unstorage.unjs.io/drivers/lru-cache) instead with [ttl option](https://www.npmjs.com/package/lru-cache#ttl).

PR #146, opened 2023-02-06
feat: `lru-cache` driver

pi0:
Related to #15, #145 and #13 (ttl)

Implement new `lru-cache` driver which is almost same as `memory` but limits the in memory cache entries with optional feature to invalidate them on a generic timeout. later we can integrate it with #13.


package.json
```diff
@@ -46,6 +46,7 @@
     "h3": "^1.1.0",
     "ioredis": "^5.3.0",
     "listhen": "^1.0.2",
+    "lru-cache": "^7.14.1",
```
pi0, line 49: Note: This is CJS but almost the best impl to use :}

PR #148, opened 2023-02-07
feat: `azure-storage-table`

itpropro:


README.md
```diff
@@ -622,6 +622,54 @@ const storage = createStorage({
 - `storage`: The name of the table to read from. It defaults to `storage`.
 - `boostCache`: Whether to enable cached queries: see [docs](https://planetscale.com/docs/concepts/query-caching-with-planetscale-boost#using-cached-queries-in-your-application).
 
+### `azure-storage-table`
```
pi0, line 625: Can you please check if this works with CF workers and browsers? (we might add (Node) if is node-js supported)

package.json
```diff
@@ -75,6 +82,7 @@
     "jsdom": "^21.1.0",
     "monaco-editor": "^0.34.1",
     "msw": "^1.0.0",
+    "npm-run-all": "^4.1.5",
```
pi0, line 85: seems unused

package.json
```diff
@@ -36,7 +36,9 @@
     "lint": "eslint --ext .ts . && prettier -c src test demo",
     "prepack": "pnpm build",
     "release": "pnpm test && changelogen --release && git push --follow-tags && pnpm publish",
-    "test": "pnpm lint && vitest run --coverage",
+    "test": "pnpm lint && run-p -r azurite-table-storage vitest",
+    "vitest": "vitest run --coverage",
+    "azurite-table-storage": "azurite --silent",
```
pi0, line 41: seems unused. (was it for local testing?)

src/drivers/azure-storage-table.ts
```diff
@@ -0,0 +1,143 @@
...
+    // fromConnectionString is only available in Node.js runtime, not in browsers
+    client = TableClient.fromConnectionString(connectionString, tableName);
+  } else {
+    const credential = new DefaultAzureCredential();
+    client = new TableClient(
+      `https://${accountName}.table.core.windows.net`,
+      tableName,
+      credential
+    );
+  }
```
pi0, line 83: I think we can lazy init `client`. See planetscale as an example.

src/drivers/azure-storage-table.ts
```diff
@@ -0,0 +1,143 @@
...
+      };
+      await client.upsertEntity(entity, "Replace");
+      return;
+    },
+    async removeItem(key) {
+      await client.deleteEntity(partitionKey, key);
+      return;
+    },
+    async getKeys() {
+      const iterator = client.listEntities().byPage({ maxPageSize: 1000 });
```
pi0, line 116: 1000 could be maybe configurable?

package.json
itpropro: `run-p` in the `test` task comes from `npm-run-all`.

README.md
itpropro: It depends on which authentication method is used, this is documented at the specific authentication method in the README and in the typedocs for the parameters. Only `DefaultAzureCredential` and `AzureSASCredential` is supported in browser.

package.json
itpropro: I'm not sure what you mean, they are all 3 used in the `test` task :) Azurite is the Azure storage emulator and is started before vitest runs to emulate table and other storage endpoints. If you think it is overhead for the test pipeline, I can remove it and try to do some mocking for that.

src/drivers/azure-storage-table.ts
itpropro: Lazy init of the client would lead to a non shared instance if used in a serverless environment that's why it's initialized at driver creation. Also, the client does not actually connect to Azure, most of the getXClient functions are not doing any network requests, so it's mostly lazy by default.

src/drivers/azure-storage-table.ts
itpropro: 1000 is just the biggest number possible to reduce the amount of network requests. If you think there is a benefit for the user, I can of course introduce a prameter for that.

package.json
pi0: Hmm right. What if we run it in a beforeAll step using execa and then kill the process after?

src/drivers/azure-storage-table.ts
pi0: I know. It is for behavior consistency for other drivers (also prevents extra sync overhead and errors on server startup, delaying anything to the first usage from driver)

src/drivers/azure-storage-table.ts
pi0: You are the pro to decide this. From an overall look, seems valid to make it configurable if the client library itself accepts it.

package.json
itpropro: I fiddled around a lot with the npm package and tried to run it with node `spawn` at first, but it's not really made to be run like this and had a lot of inconsistent issues and problems being started and killed properly. I can try to mock the replies, as I would like to avoid spawning it from node.

src/drivers/azure-storage-table.ts
itpropro: Makes sense, I will make it lazy to have a consistent implementation.

README.md
itpropro: CF workers is something I have to test, haven't used the so far, just local workers.

package.json
itpropro: Got it to work with exec

README.md
itpropro: @pi0 I would like to merge and validate CF functionality for all of the Azure libraries in the following days when I have finished getting comfortable with CF.

itpropro: The Azure storage SDKs don't seem to be compatible with CF workers currently, especially when using @azure/identity to login. I added a note for that to the README.

README.md
```diff
@@ -622,6 +622,56 @@ const storage = createStorage({
...
+Please make sure that the table you want to use exists in your storage account.
+
+```js
+import { createStorage } from "unstorage";
+import azureStorageTableDriver from "unstorage/drivers/azure-storage-table";
+
+const storage = createStorage({
+    driver: azureStorageTableDriver({
+      accountName: "myazurestorageaccount",
+    }),
```
danielroe, line 653:
```suggestion
  driver: azureStorageTableDriver({
    accountName: "myazurestorageaccount",
  }),
```

README.md
```diff
@@ -622,6 +622,56 @@ const storage = createStorage({
...
+});
+```
+
+**Authentication:**
+
+The driver supports the following authentication methods:
+
+- **`DefaultAzureCredential`**: This is the recommended way to authenticate. It will use managed identity or environment variables to authenticate the request. It will also work in a local environment by trying to use Azure CLI or Azure PowerShell to authenticate. <br>
+⚠️ Make sure that your Managed Identity or personal account has the `Storage Table Data Contributor` role assigned to it, even if you already are `Contributor` or `Owner` on the storage account.
```
danielroe, line 662:
```suggestion
- **`DefaultAzureCredential`**: This is the recommended way to authenticate. It will use managed identity or environment variables to authenticate the request. It will also work in a local environment by trying to use Azure CLI or Azure PowerShell to authenticate.

   ⚠️ Make sure that your Managed Identity or personal account has the `Storage Table Data Contributor` role assigned to it, even if you already are `Contributor` or `Owner` on the storage account.
```

danielroe (approved):

pi0 (approved):
Thanks!

PR #150, opened 2023-02-10
redisDriver adds the setKeyExpire method to set the expiration time of the key. After the key expires, it will no longer be available. The unit is in seconds.

rdzhaoxin:
### `storage.setKeyExpire(key, seconds)`

Set the expiration time of the key. After the key expires, it will no longer be available. The unit is in seconds.

**Note:** Only for redisDriver.

```js
await storage.setKeyExpire("foo:bar", 1); // 1s
```

Test file path: test/drivers/redis.test.ts.

Before testing, you need to modify the ip, password, and port in the file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,6 +101,16 @@ If value is `undefined`, it is same as calling `removeItem(key)`.
 await storage.setItem("foo:bar", "baz");
 ```
 
+### `storage.setKeyExpire(key, seconds)`
+
+Set the expiration time of the key. After the key expires, it will no longer be available. The unit is in seconds.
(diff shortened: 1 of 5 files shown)


rd-carter:
@rbndelrio @manniL @benjamincanac @Atinux

pi0:
Thanks for the PR dear @rdzhaoxin (please keep doing it again!). Supporting ttl is a good idea however it is driver dependent.

Please see my comment at https://github.com/unjs/unstorage/issues/13#issuecomment-1448906670

PR #154, opened 2023-02-19
feat: `azure-storage-blob` driver

itpropro:


src/drivers/azure-storage-blob.ts
```diff
@@ -0,0 +1,180 @@
...
+    },
+    async setItem(key, value) {
+      await getContainerClient()
+        .getBlockBlobClient(key)
+        .upload(value, Buffer.byteLength(value));
+      return;
+    },
+    async removeItem(key) {
+      await getContainerClient().getBlockBlobClient(key).delete();
+      return;
```
danielroe, line 109:
```suggestion
```

src/drivers/azure-storage-blob.ts
```diff
@@ -0,0 +1,180 @@
...
+        return (await streamToBuffer(blob.readableStreamBody)).toString();
+      } catch {
+        return null;
+      }
+    },
+    async setItem(key, value) {
+      await getContainerClient()
+        .getBlockBlobClient(key)
+        .upload(value, Buffer.byteLength(value));
+      return;
```
danielroe, line 105:
```suggestion
```

src/drivers/azure-storage-blob.ts
```diff
@@ -0,0 +1,180 @@
...
+        await Promise.all(
+          page.segment.blobItems.map(
+            async (blob) =>
+              await getContainerClient().deleteBlob(blob.name, {
+                deleteSnapshots: "include",
+              })
+          )
+        );
+      }
+      return;
```
danielroe, line 147:
```suggestion
```

danielroe (approved):

pi0 (approved):
Nice work & LGTM. I think next steps we can enable raw support for this driver (which is probably it's main usesecause i guess)

itpropro:
> Nice work & LGTM. I think next steps we can enable raw support for this driver (which is probably it's main usesecause i guess)

Yes, raw support would be a great next step and the driver would be perfectly fitted for that :)

PR #156, opened 2023-02-19
feat: `azure-app-configuration` driver

itpropro:


itpropro:
> (not tested. trusting works as expected)

Testing will only work against a production instance, basically the same as with the cosmos driver. I tested all of the drivers against free and paid actual instances of the services.
Will also create a functional mock, in addition we should create a free instance of the app config in the shared Azure environment as well.

PR #158, opened 2023-02-20
feat(driver):  Added Azure CosmosDB driver

itpropro:
Added driver for Azure CosmosDB NoSQL API. For other APIs like MongoDB, please use the native API driver.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ tmp
 /drivers
 /server*
 /test.*
+__*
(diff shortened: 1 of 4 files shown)


pi0 (approved):

pi0:
Thanks. I think we need to implement tests by running mock server? Disabled for now trusting you base functionality works as expected (also made a small untested change in queries to select only fields we need i hope it is fine)

itpropro:
> Thanks. I think we need to implement tests by running mock server? Disabled for now trusting you base functionality works as expected (also made a small untested change in queries to select only fields we need i hope it is fine)

I tested against a production cosmos db, as there is no suitable emulator and no really decent mock that would actually verify functionality available. Maybe we can just setup a Key Vault and a CosmosDB in Free/Serverless tier for testing in the same tenant where the "nitro-deploys" resources are already located. What do you think @danielroe

pi0:
Then we might only mock package interface to make sure it calls expected methods:)

itpropro:
> Then we might only mock package interface to make sure it calls expected methods:)

I will prepare that :)

PR #159, opened 2023-02-20
feat(driver): Added Azure Kay Vault driver

itpropro:


README.md
```diff
@@ -622,6 +622,52 @@ const storage = createStorage({
...
+
+```js
+import { createStorage } from "unstorage";
+import azureKeyVault from "unstorage/drivers/azure-key-vault";
+
+const storage = createStorage({
+    driver: azureKeyVault({
+      vaultName: "testunstoragevault",
+  }),
+});
```
danielroe, line 655:
```suggestion
  driver: azureKeyVault({
    vaultName: "testunstoragevault",
  }),
});
```

README.md
```diff
@@ -622,6 +622,52 @@ const storage = createStorage({
...
+```
+
+**Authentication:**
+
+The driver supports the following authentication methods:
+
+- **`DefaultAzureCredential`**: This is the recommended way to authenticate. It will use managed identity or environment variables to authenticate the request. It will also work in a local environment by trying to use Azure CLI or Azure PowerShell to authenticate. <br>
+⚠️ Make sure that your Managed Identity or personal account has either the `Key Vault Secrets Officer` RBAC role assigned or being member of an access policy that grants `Get`, `List`, `Set`, `Delete` and `Purge` secret permissions.
```
danielroe, line 663:
```suggestion
- **`DefaultAzureCredential`**: This is the recommended way to authenticate. It will use managed identity or environment variables to authenticate the request. It will also work in a local environment by trying to use Azure CLI or Azure PowerShell to authenticate.

   ⚠️ Make sure that your Managed Identity or personal account has either the `Key Vault Secrets Officer` RBAC role assigned or is a member of an access policy that grants `Get`, `List`, `Set`, `Delete` and `Purge` secret permissions.
```

src/drivers/azure-key-vault.ts
```diff
@@ -0,0 +1,128 @@
...
+      try {
+        const secret = await getKeyVaultClient().getSecret(encode(key));
+        return secret.value;
+      } catch {
+        return null;
+      }
+    },
+    async setItem(key, value) {
+      await getKeyVaultClient().setSecret(encode(key), value);
+      return;
```
danielroe, line 59:
```suggestion
```

src/drivers/azure-key-vault.ts
```diff
@@ -0,0 +1,128 @@
...
+    },
+    async removeItem(key) {
+      const poller = await getKeyVaultClient().beginDeleteSecret(encode(key));
+      await poller.pollUntilDone();
+      await getKeyVaultClient().purgeDeletedSecret(encode(key));
+      return;
```
danielroe, line 65:
```suggestion
```

danielroe (commented):
looks good to me! what about something like https://github.com/peveuve/ms-vault-mock for testing?

itpropro:
> looks good to me! what about something like https://github.com/peveuve/ms-vault-mock for testing?

Great find, I will check that. The main problem would be that it hides the actual slowness of Azure Kay Vault, but for unit testing it great.

itpropro:
> looks good to me! what about something like https://github.com/peveuve/ms-vault-mock for testing?

I tested that and it's quite complex to get it to run with the current version of Key Vault packages, as you have to generate cert pairs, disable HTTPS validation and create some workaround due so recent API changes. The package wasn't updated for two years, which means there are a lot of old and complex dependencies from lodash to express and the Key Vault API itself is also old and has no support for purge and soft delete.
I would prefer a free Key Vault hosted instance tbh @danielroe

pi0 (approved):
Thanks (also skipping tests for now. trusting it works as expected)

itpropro:
> Thanks (also skipping tests for now. trusting it works as expected)

The only thing about this driver (different to my other PRs) is that Azure Key Vault is a very slow service. The clear action can take a minute, although I tried to make it at least a little better with concurrency. 
It’s also in the README, but if you don’t like it, because of the actual service slowness, we can revisit it.

pi0:
Slowness is okay and i guess somehow expected for encryption. As long as we can test locally we can add some additional tests only running in CI.

PR #160, opened 2023-02-21
feat(driver):  Refactored redis driver

itpropro:
- Refactored redis driver to conform with unstorage driver standards
- Added tests with in-memory redis
- Added cluster support, fixes #157
- Updated docs

PR #163, opened 2023-02-23
fix(redis): fix clear method

### LINKED ISSUE #162: cannot clear the cache with redis driver
### Environment

node v19

### Reproduction

```
import { createStorage } from 'unstorage'
import redisDriver from 'unstorage/drivers/redis'

  driver: redisDriver({ base: 'test:' }),
})

await storage.setItem('item', 'val')

await storage.clear()

// item still exist
const item = await storage.getItem('item', 'val')
```

after runing the above code, the item is still exist in redis

```
import { createStorage } from 'unstorage'
import redisDriver from 'unstorage/drivers/redis'

// no items have been set yet...
await storage.clear()
```

if there is no items in redis initially, the clear mehtod will throw error

```ReplyError: ERR wrong number of arguments for 'del' command```

### Describe the bug

cannot clear the cache with redis driver

### Additional context

_No response_

### Logs

_No response_

jamwong-ecosa:
fix #162

### DIFF
--- a/src/drivers/redis.ts
+++ b/src/drivers/redis.ts
@@ -35,7 +35,7 @@ export default defineDriver<RedisOptions>((_opts) => {
     async clear() {
       const keys = await redis.keys(r("*"));
       if (keys.length === 0) {
-        return
+        return;
       }
       return redis.del(keys).then(() => {});
     },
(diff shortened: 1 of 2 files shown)


danielroe (approved):
Thanks!

Enabled unit test on redis as well: https://github.com/unjs/unstorage/pull/164. Once that's enabled the test I've added here should prevent regression on this bug.

itpropro:
Should also be fixed in #160 @danielroe 😃

pi0 (approved):
Thanks!

PR #164, opened 2023-02-23
test: add unit test for redis driver

danielroe:
Following https://github.com/unjs/unstorage/pull/163, this enables unit testing with redis. We can then update the `testDriver` utility in that PR to add a regression test for the bug it is fixing.


itpropro:
He @danielroe,
I already added tests, including different authentication methods to #160. They are not using a mock, but instead a in-memory actual redis implementation like the in-memory mongodb one at the mongodb driver.

danielroe:
Rats, how did I not see this?! That sounds better. I'll check later on.

pi0 (approved):
Thanks!

PR #179, opened 2023-03-08
feat(drivers): add session storage driver

CorentinTh:
Hi !

This PR adds the [session storage](https://developer.mozilla.org/en-US/docs/Web/API/Window/sessionStorage) driver

- [x] Driver implementation
- [x] Unit test
- [x] Updated doc

### DIFF
--- a/docs/content/6.drivers/session-storage.md
+++ b/docs/content/6.drivers/session-storage.md
@@ -1,0 +1,22 @@
+---
+navigation.title: Session Storage
+---
+
+# Session Storage
+
+Store data in [sessionStorage](https://developer.mozilla.org/en-US/docs/Web/API/Window/sessionStorage).
+
+```js
+import { createStorage } from "unstorage";
+import sessionStorageDriver from "unstorage/drivers/session-storage";
+
+const storage = createStorage({
+  driver: sessionStorageDriver({ base: "app:" }),
+});
+```
+
+**Options:**
+
+- `base`: Add `${base}:` to all keys to avoid collision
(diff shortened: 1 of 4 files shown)


CorentinTh:
It has a similar api to the local storage driver, so one could have use :
```js
  driver: localStorageDriver({ base: "app:", localStorage: window.sessionStorage }),
});
```
So what do you think of this driver ?

CorentinTh:
It could also be written like :

CorentinTh:
Since they have a similar API with localstorage, it can also be written like :

```js
import { defineDriver } from "./utils";
import localStorageDriver from './localstorage'

export interface SessionStorageOptions {
  base?: string;
  window?: typeof window;
  sessionStorage?: typeof window.sessionStorage;
}

export default defineDriver((opts: SessionStorageOptions = {}) => {
  if (!opts.window) {
    opts.window = typeof window !== "undefined" ? window : undefined;
  }
  if (!opts.sessionStorage) {
    opts.sessionStorage = opts.window?.sessionStorage;
  }
  if (!opts.sessionStorage) {
    throw new Error("localStorage not available");
  }

 return localStorageDriver({...opts, localStorage: opts.sessionStorage})
});
```
But it induces coupling between drivers, and decrease maintainability 🤷🏻

CorentinTh:
Hi @pi0 !
Let me know if I have something to change 🙏🏻

pi0 (approved):
LGTM 💯 

Regarding refactor to extend local storage, let's see how frequently we need to update both and if makes sense to extend for easier maintenance.

Issue #189, opened 2023-03-23
Duplicate line


ManUtopiK:
https://github.com/unjs/unstorage/blob/cca7ebf49fbbc12640af675d5c6ef4e8a7f24ad9/src/index.ts#L17-L19

thetejasmagade:
Hey @pi0, Should I update?

pi0:
Sure feel free to make a pr

NozomuIkuta:
Let me close this issue since the PR has been already merged.

PR #190, opened 2023-03-25
fix: removed duplicate line

thetejasmagade:
Solves: #189 

Removed Duplicate Line 
https://github.com/unjs/unstorage/blob/cca7ebf49fbbc12640af675d5c6ef4e8a7f24ad9/src/index.ts#L17-L19


pi0 (approved):
Thanks!

PR #196, opened 2023-04-09
fix(github): optional properties

davestewart:
The supposedly optional driver properties aren't optional on the interface, which makes configuration more work than it needs to be.

This PR makes `branch`, `dir` and `ttl` optional, as they have defaults.


pi0 (approved):
Thanks!

PR #200, opened 2023-04-11
fix(planetscale): fix `hasItem`

winton:


src/drivers/planetscale.ts
```diff
@@ -50,10 +50,10 @@ export default defineDriver((opts: PlanetscaleDriverOptions = {}) => {
     options: opts,
     hasItem: async (key) => {
       const res = await getConnection().execute(
-        `SELECT COUNT(id) from ${opts.table} WHERE id=:key;`,
+        `SELECT COUNT(id) as value from ${opts.table} WHERE id=:key;`,
```
danielroe, line 53: what was the issue you were encountering with the previous code?

However looking at this, what about this instead?

```suggestion
        `SELECT EXISTS (SELECT 1 FROM ${opts.table} WHERE id = :key);`,
```

winton: The issue is that `hasItem` always returns true because the query always returns a row.

The code you've suggested results in these rows:

```
rows: [
    {
      "exists (select 1 from users where id = 'xxx' limit 1)": '0'
    }
  ]
```

Then `res.size >= 0` still returns true despite there being no matches.

winton: Sure, I've changed the code to use `SELECT EXISTS`. I made the `rows` function generic which so I can reuse it for the rows result.

danielroe (approved):

pi0:
Thanks!

PR #206, opened 2023-04-20
refactor!: return `undefined` instead of `null` for non existing keys

ur92:


CHANGELOG.md
pi0, line 1: Please revert changes to this file

package.json
pi0, line 1: Please revert version bump in PR

pi0:
@ur92 Since the major change in unstorage requires ecosystem adoption (nitro and nuxt mainly) i am delaying this change.

Also remembered about [LocalStorage `getItem`](https://developer.mozilla.org/en-US/docs/Web/API/Storage/getItem#return_value) behavior returning `null` (localstorage interface design was a main inspiration for unstorage). 

I think this change would be better to waited and discussed little more before happening.

pi0:
Reworking in #260 to resolve merge conflicts. Once again thanks for working on it ❤️

PR #213, opened 2023-05-02
feat: add `vercel-kv` driver

danielroe:
This adds a new driver (based on the `redis` driver) for the Vercel KV store.

See https://vercel.com/docs/storage/vercel-kv for an implementation

For a demo, see https://github.com/danielroe/vercel-kv, and you can hit https://vercel-kv.vercel.app/api/kv for a response rendered using the KV store.


src/drivers/vercel-kv.ts
```diff
@@ -0,0 +1,43 @@
...
+import type { RedisConfigNodejs } from "@upstash/redis";
+
+import { defineDriver } from "./utils";
+
+export interface RedisOptions extends RedisConfigNodejs {
+  base?: string;
+}
+
+export default defineDriver<RedisOptions>((opts) => {
+  const redis = createClient(opts);
```
pi0, line 11: Would you please make changes to make client lazy initialized? 🙏

src/drivers/vercel-kv.ts
```diff
@@ -0,0 +1,43 @@
+import { createClient } from "@vercel/kv";
+import type { RedisConfigNodejs } from "@upstash/redis";
```
pi0, line 2: As i understand, this is a type-only dependency for us. Shall we instead inline the config types?

src/drivers/vercel-kv.ts
```diff
@@ -0,0 +1,43 @@
...
+    getItem(key) {
+      return redis.get(r(key));
+    },
+    setItem(key, value) {
+      return redis.set(r(key), value).then(() => {});
+    },
+    removeItem(key) {
+      return redis.del(r(key)).then(() => {});
+    },
+    getKeys() {
```
pi0, line 32: Perf:We could implement getKeys with base while now it always makes a query for all keys and filters on client side

src/drivers/vercel-kv.ts
```diff
@@ -0,0 +1,43 @@
...
+      return redis.keys(r("*"));
+    },
+    async clear() {
+      const keys = await redis.keys(r("*"));
```
pi0, line 36: Keys is not supported here as well

src/drivers/vercel-kv.ts
danielroe: this is a type-level dependency for _us_ but a dependency of `@vercel/kv`. I'll add it back in as an optional peer dep. I think as it is installed anyway, it would be better to use its types.

pi0: I guess for type support for pnpm users without shamefully-hoist flag they need to install both then right?

danielroe: yes, they would 👍

src/drivers/vercel-kv.ts
danielroe: what do you mean?

danielroe:
great idea!

src/drivers/vercel-kv.ts
pi0: Yes. And yes other driver needs fix

src/drivers/vercel-kv.ts
danielroe: no, not too much trouble at all

Intevel (approved):
Looks great - Good work on this 👯

Intevel:
@danielroe What do you think about adding this to the Nuxt 3 Documentation? To the Vercel Deployment?

pi0:
Driver tests to be added later (as well as upstash provider. they are almost identical!)

pi0:
Followup https://github.com/unjs/unstorage/issues/218

PR #227, opened 2023-05-18
feat(drivers): add support for Deno Kv

### LINKED ISSUE #224: Support deno kv
### Describe the feature

Deno has a built-in kv storage, it would be great if we can use it with unstorage!

### Additional information

so1ve:
close #224

I'm not sure about how to test it ...

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,34 @@
 
 All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
 
+## v1.6.1
+
+[compare changes](https://github.com/unjs/unstorage/compare/v1.6.0...v1.6.1)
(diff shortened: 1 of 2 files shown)


pi0:
Thanks for the PR 🚀 

Do you mind to merge types and util to the same driver util? (to make project files less sparse)

For testing, we might make a mock interface that is similar to Deno KV. (alternative would be to setup deno and use a HTTP driver but seems overwork for)

so1ve:
> Do you mind to merge types and util to the same driver util?

Done!

> For testing...

How about using deno's native testing?

so1ve:
...WTF where are my changes

so1ve:
??????!!!!!!!! pull deleted my changes??????

so1ve:
I think in need to create a new branch instead of main :( bad habits

PR #232, opened 2023-05-21
fix: fixed #222

### LINKED ISSUE #222: getItemRaw() on PNG returns null
### Environment

Node: 18.6.0
NPM: 8.19.2
unjs/unstorage: 1.6.0

### Reproduction

Hopefully description is enough here

### Describe the bug

On a brand new Nuxt 3 project I'm using the Nitro Custom Server Assets feature to access files from public folder on server-side API. To access one of the files I'm using the `getItemRaw()` method expecting the buffer as a result but getting `null` instead.

```
// Nitro config
serverAssets: [{
  baseName: 'public',
  dir: '../public'
}]

// Server-side API
const imageBuffer = await useStorage('assets:public').getItemRaw('x.png')
console.log(imageBuffer) // <--- returns null
```

### Additional context

Looks like driver is `memory` when used with Nuxt. Could this be an issue?
I also described the problem here: https://github.com/unjs/nitro/discussions/1230

### Logs

_No response_

maou-shonen:
fixed #222
The issue is that `prefixStorage` is not initializing `getItemRaw` and `setItemRaw`.

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -5,7 +5,9 @@ type StorageKeys = Array<keyof Storage>;
 const storageKeyProperties: StorageKeys = [
   "hasItem",
   "getItem",
+  "getItemRaw",
   "setItem",
+  "setItemRaw",
   "removeItem",
   "getMeta",
   "setMeta",


pi0:
Thanks!

PR #234, opened 2023-05-25
feat(drivers): add support for AWS DynamoDB

daaru00:
Add support for [Amazon DynamoDB](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Introduction.html).
Also add support for [DynamoDB Time to Live (TTL)](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/howitworks-ttl.html), this is order to manage key expiration.

A note about "aws-dynamodb" driver's tests: they will fail if no valid [AWS credentials](https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide/setting-credentials-node.html) will be found (using profile or environment variables).

### DIFF
--- a/src/drivers/aws-dynamodb.ts
+++ b/src/drivers/aws-dynamodb.ts
@@ -1,4 +1,4 @@
-import { createRequiredError, defineDriver } from "./utils";
+import { createError, createRequiredError, defineDriver } from "./utils";
 import { DynamoDBClient } from "@aws-sdk/client-dynamodb";
 import {
   DynamoDBDocumentClient,
@@ -15,7 +15,9 @@ export interface DynamoDBStorageOptions {
   attributes?: {
     key?: string;
     value?: string;
+    ttl?: string;
   };
+  expireIn?: number;
 }
 
 const DRIVER_NAME = "aws-dynamodb";
@@ -29,6 +31,7 @@ export default defineDriver((opts: DynamoDBStorageOptions) => {
     opts.attributes = {
       key: "key",
       value: "value",
+      ttl: "ttl",
     };
   } else {
     if (!opts.attributes.key) {
@@ -38,6 +41,16 @@ export default defineDriver((opts: DynamoDBStorageOptions) => {
     if (!opts.attributes.value) {
       opts.attributes.value = "value";
     }
+
+    if (!opts.attributes.ttl) {
+      opts.attributes.value = "ttl";
+    }
+  }
+
+  opts.expireIn =
+    opts.expireIn === undefined ? 0 : parseInt(`${opts.expireIn}`);
+  if (Number.isNaN(opts.expireIn) || opts.expireIn < 0) {
+    throw createError(DRIVER_NAME, "expireIn");
   }
 
   let client;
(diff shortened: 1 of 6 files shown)


pi0:
Thanks for the PR @daaru00. Checking locally, the types and test are failing. Can you please check? 🙏🏼

daaru00:
Hello @pi0,

May I ask what error message you get? 
I suspect it's something related to AWS credentials, without them the integration tests will fail not being able to access a valid AWS account (in order to create a temporary table to run tests on).

pi0:
> I suspect it's something related to AWS credentials, without them the integration tests will fail not being able to access a valid AWS account (in order to create a temporary table to run tests on).

Do you think we can mock AWS library to make sure tests run locally?

daaru00:
@pi0 I did a test using the official mock library managed by AWS: https://aws.amazon.com/it/blogs/developer/mocking-modular-aws-sdk-for-javascript-v3-in-unit-tests/. 
Actually I replaced the api calls using Map saving key and value in-memory:
```js
client
    .on(GetCommand)
    .callsFake(input => {
      const key = input.Key.key
      if (!data.has(key)) {
        return { Item: undefined }
      }
      return { Item: { key, value: data.get(key) } }
    })
    .on(ScanCommand)
    .callsFake(() => {
      return { Items: Array.from(data.entries()).map(([ key, value ]) => ({ key, value })) }
    })
    .on(PutCommand)
    .callsFake(input => {
      data.set(input.Item.key, input.Item.value)
    })
    .on(DeleteCommand)
    .callsFake(input => {
      data.delete(input.Key.key)
    })
```
Is this the expected result?

ps: I also had to update the AWS SDK due to some internal conflicts with typescript types

src/drivers/aws-dynamodb.ts
```diff
@@ -0,0 +1,161 @@
...
+  table: string;
+  region?: string;
+  credentials?: any;
+  attributes?: {
+    key?: string;
+    value?: string;
+    ttl?: string;
+  };
+  expireIn?: number;
+  client?: DynamoDBDocumentClient;
```
Hebilicious, line 21: Would be nice to have ttl as a top level option here, for consistenct with other drivers #262

src/drivers/aws-dynamodb.ts
```diff
@@ -0,0 +1,161 @@
...
+    name: DRIVER_NAME,
+    options: opts,
+    async hasItem(key) {
+      const item = await getItemValue(key);
+      return !!item;
+    },
+    async getItem(key) {
+      return await getItemValue(key);
+    },
+    async setItem(key, value) {
```
Hebilicious, line 147: We could accept, key, value, options with { ttl } in options for per object expiration.

aheissenberger:
what is the timeline for this to be merged into master?

robertsosinski:
Pinging this PR as its been a year since last comment. Would be great to get this into unstorage.

Thanks!

robokozo:
Have you given any thought to setting both the pk and the sk? maybe that goes beyond the scope of the unstorage use case though. i was just giving it some thought since i saw the 'scan' being used to get the keys and was thinking about someone trying to use this in an existing table (single table design). You'd probably want a pk with a fixed value, like pk: "unstorageKV" and the sk would be your "actual" key. maybe this a prefix storage thing?

Issue #226, opened 2023-05-17
Add cloudflare R2 driver 


pperzyna:
### Describe the feature

It would be useful to add R2 cloudflare support 

### Additional information

- [ ] Would you be willing to help implement this feature?

PR #235, opened 2023-05-26
feat: `cloudflare-r2-binding` driver

pi0:
resolves https://github.com/unjs/unstorage/issues/226

Support for [cloudflare r2](https://developers.cloudflare.com/r2/api/workers/workers-api-usage/) (within workers). Support for HTTP / S3 compatible driver coming soon for other envs.

TODO:
- [ ] ~~Try to use miniflare v3 instead of mocks~~ delayed
- [ ] Test key normalization to convert `:` <> `/`


pi0:
Moving forward as experimental driver

Issue #32, opened 2021-11-05
Use Cloudflare KV native meta


pi0:
https://developers.cloudflare.com/workers/runtime-apis/kv#metadata

By using native meta, we can allow expiration. (depends on #13)

Hebilicious:
We should be able to grab the 3rd parameter from `.setItem` and pass it to cloudflare KV :

- For [HTTP single](https://developers.cloudflare.com/api/operations/workers-kv-namespace-write-key-value-pair-with-metadata)
- For [HTTP batch](https://developers.cloudflare.com/api/operations/workers-kv-namespace-write-multiple-key-value-pairs)

The single version is poorly documented, but I assume it works if you pass the same options in the body. Otherwise we can just modify the HTTP driver to use the batch API.

```json
  {
    "base64": false,
    "expiration": 1578435000,
    "expiration_ttl": 300,
    "key": "My-Key",
    "metadata": {
      "someMetadataKey": "someMetadataValue"
    },
    "value": "Some string"
  }
  ```
  
  ... ie we should be able to do : 
  
  ```ts
  await storage.setItem("My-Key", "Some string", { expiration: 1578435000, expiration_ttl: 300, base64: false, metadata: {someMetadataKey: "someMetadataValue"}} );
  ```
  and get the query above sent properly.

matfire:
I've just started working on a project using unstorage and kind of need this feature. If nobody is working on it, can work on implementing this ?
I'd rather modify it to use the batch endpoint since it handles expiration_ttl directly. If using the single put method, we'd still need to do a request before that to check the metadata for an expiration date, right ?

The only issue I can see in this is that the naming convention for the binding and http API is not the same (`expirationTtl` for binding, `expiration_ttl` for the API), so we'd probably need to abstract it and convert it back to right format (i.e. check for a parameter called `ttl` and handle that internally)

Please let me know if I can help :)

matfire:
Hello, would love to work on this if nothing is currently happening

PR #236, opened 2023-05-26
feat: support cloudflare options

Hebilicious:
fix https://github.com/unjs/unstorage/issues/32

This PR allows the following : 

```ts
await storage.setItem("key", "value", 
{ 
  expiration: 1578435000, 
  expiration_ttl: 300, // http driver
  expirationTtl: 300, // binding driver
  base64: false,
  metadata: { someMetadataKey: "someMetadataValue" }
});
```

- implements the cloudflare options such as metadata, expiring keys, etc. 
	- Forward the 3rd argument (options) to the `cloudflare-kv-binding`and `cloudflare-kv-http` drivers `setItem` method.
	- Changes the put implementation of the `cloudflare-kv-http` to use the bulk API ([to support all options](https://developers.cloudflare.com/api/operations/workers-kv-namespace-write-multiple-key-value-pairs)). 
- Refactor the `cloudflare-kv-http` tests to use the lastest version of `msw`

@pi0 I have some concerns, I'll list them here.

## Issues with msw and ofetch
I upgraded to the latest msw because the tests were disabled.  However I found out that there's a bug with `ofetch` and `msw` that doesn't happen with native fetch :/  🤔 For now I changed the logic to use ofetch.native and fetch for the tests only, I'm open to suggestion to make this more elegant. From the top of my head we could : 

- Extend this idea by allowing the user to specify a fetch function.

```ts
  driver: cloudflareKVHTTPDriver({
   //...
    fetch: myFetch
  }),
});
```
- Get rid of msw and handle the test logic differently
- Find a way to make msw and oftech work together

## kv-binding-test

I've noticed that the kv-binding-tests are using the memory driver. Should we use [miniflare kv](https://miniflare.dev/storage/kv) instead ? 

## Options "API design"

For now I'm leverage the 3rd argument of the setItem method, but I was thinking that we could require users to pass them nested under a key (`{ cloudflare: {} }`), or use 4th argument, in-case we want to keep the 3rd argument open for unstorage specific options... Thoughts ?

## Get options and metadata

It's not possible for now to retrieve metadata or key expiration values with the default drivers:
Cloudflare KV get accepts options like this, see [docs](https://developers.cloudflare.com/workers/runtime-apis/kv/#listing-keys) :

```ts
NAMESPACE.get(key, { cacheTtl: 3600 });
```
and to retrieve metadata : 

```ts
const { value, metadata } = await NAMESPACE.getWithMetadata(key, { cacheTtl: 3600 });
// or the list method
NAMESPACE.list({prefix?: string, limit?: number, cursor?: string})
```

Should we do something like this with unstorage ? 

```ts
await storage.getItem("key", { cacheTtl: 3600 }); // pass as is
await storage.getItem("key", { withMetadata: true }); // detect and use a different method.
// Object prefix style (we can provide type safety)
await storage.getItem("key", {cloudflare: {options:  { cacheTtl: 3600 }, metadata: false }});
// 4th argument (3rd one reserved for unstorage)
await storage.getItem("key", {},{cloudflare: {options:  { cacheTtl: 3600 }, metadata: false }});
```

Or maybe this is out of scope?


McPizza0:
@pi0 you recently added TTL to Redis driver
Can you review this please?

Hebilicious:
Closing in favor of https://github.com/unjs/unstorage/pull/255

PR #240, opened 2023-05-31
feat: experimental operation batching

### LINKED ISSUE #198: support `setItems` and `getItems` for operation batching
### Describe the feature

I'm currently building a driver for gitlab. #199 
I just forked the github driver, write tests, adapted `fetchFiles` and `getItem` for gitlab api, and done !

What I want to do is to provide a feature to `setItem` with a `create/commit` http post request.
But I don't want to commit each files one by one, I want to commit a bunch of files, with a commit message passed in `opts` and update the store internally with the http response.

A `setItems` function could be useful to do that. I think it could also be implemented for the github driver.
 

### Additional information

Hebilicious:
Resolves #198

## Description 

Add the following APIs :

- setItems()
- getItems() 

The getItems() API supports 2 syntaxes for setting metadata on multiple keys.
In future PRs we can individually implement more efficient batching for drivers that have a true batching API. There's the localstorage driver that has a getItems implemented has a proof-of-concept, but it can be removed as it's doing the same as the fallback. 

@pi0 Note that the `getItems` method will retrieve the first mountpoint. Something we could do is to use Promise.any on all the mountpoints to get the first one that succeed. I'm not sure if that's desirable as it will make un-needed reads... But maybe that's what is expected with multiple mountpoints? 

TODO :
- [x] Base implementation
- [x] Tests
- [ ] Docs


src/drivers/localstorage.ts
```diff
@@ -32,6 +37,21 @@ export default defineDriver((opts: LocalStorageOptions = {}) => {
     getItem(key) {
       return opts.localStorage!.getItem(r(key));
     },
+    getItems(items) {
```
pi0, line 40: This can be removed

src/types.ts
```diff
@@ -23,6 +23,9 @@ export interface Driver {
     opts?: TransactionOptions
   ) => MaybePromise<StorageValue>;
   /** @experimental */
+  getItems?: (
+    items: { key: string; opts?: TransactionOptions }[] | string[]
```
pi0, line 27: We might support one opts as second param for all operations

pi0:
Update: Just pushed a refactor with a shared internal `runBatch` utility that is able to run multi level parallel (per-driver batching support when possible) in addition to some minor refactors.

pi0:
@Hebilicious Something regarding unstorage API design:

User => (not normalized input) => Storage => (normalized input) => Driver => (not normalized output) => Storage => (normalized output)

This way we make sure both mistakes and not normalized (non promises, bad keys, etc) by Users and Driver authors are gracefully handled by Storage layer.

pi0 (approved):

PR #255, opened 2023-06-30
feat: support cloudflare options with setItem

Hebilicious:
### 🔗 Linked issue

Resolves https://github.com/unjs/unstorage/issues/32

First part of https://github.com/unjs/unstorage/pull/236

### ❓ Type of change

- [x] 📖 Documentation (updates to the documentation, readme or JSdoc annotations)

### 📚 Description

This PR allows the ability to pass options to the setItem method while using the cloudflare driver : 

```ts
await storage.setItem("key", "value", 
{ 
  expiration: 1578435000, 
  expiration_ttl: 300, // http driver
  expirationTtl: 300, // binding driver
  base64: false,
  metadata: { someMetadataKey: "someMetadataValue" }
});
```

- implements the cloudflare options such as metadata, expiring keys, etc. 
	- Forward the 3rd argument (options) to the `cloudflare-kv-binding`and `cloudflare-kv-http` drivers `setItem` method.
	- Changes the put implementation of the `cloudflare-kv-http` to use the bulk API ([to support all options](https://developers.cloudflare.com/api/operations/workers-kv-namespace-write-multiple-key-value-pairs)). 

## Options "API design"

For now I'm leverage the 3rd argument of the setItem method, but I was thinking that we could require users to pass them nested under a key (`{ cloudflare: {} }`), or use a 4th argument, in-case we want to keep the 3rd argument open for unstorage specific options... Thoughts ?

### 📝 Checklist

### DIFF
--- a/docs/content/6.drivers/cloudflare-kv-binding.md
+++ b/docs/content/6.drivers/cloudflare-kv-binding.md
@@ -41,8 +41,11 @@ Refer to the [cloudflare KV docs](https://developers.cloudflare.com/workers/runt
 
 ```ts
 await storage.setItem("key", "value", {
-  expiration: 1578435000,
-  expirationTtl: 300,
-  metadata: { someMetadataKey: "someMetadataValue" },
+  ttl: 300, // key expiration shorthand (in seconds)
+  cloudflareKvBinding: {
+    expiration: 1578435000,
+    expirationTtl: 300,
+    metadata: { someMetadataKey: "someMetadataValue" },
+  },
 });
 ```
--- a/docs/content/6.drivers/cloudflare-kv-binding.md
+++ b/docs/content/6.drivers/cloudflare-kv-binding.md
@@ -33,3 +33,16 @@ const storage = createStorage({
 
 - `binding`: KV binding or name of namespace. Default is `STORAGE`.
 - `base`: Adds prefix to all stored keys
+
+## Using Cloudlflare KV Options
+
+You can pass options to cloudflare such as metadata and expiration as the 3rd argument of the `setItem` function.
+Refer to the [cloudflare KV docs](https://developers.cloudflare.com/workers/runtime-apis/kv/#writing-key-value-pairs) for the list of supported options.
+
+```ts
+await storage.setItem("key", "value", {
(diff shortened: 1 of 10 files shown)


Hebilicious:
Refactor with namespace

Hebilicious:
Added a generic ttl parameters, as well as a default ttl configurable in the options. 
Also added namespaced options under the kebabcased driver name :  `cloudflareKvBinding` and `cloudflareKvHttp`. 
Since they are different we can't use the same key.

pi0:
> Refactor TransactionOptions into 2 new types that are more explicit, SetItemOptions and GetItemOptions, and have the generic ttl?: number, and fix the type bugs without changing behaviours.

As i mentioned in other PR, it would make sense to support one kind of options interface (even if some option keys are not used by read operation). Particulary for `ttl` that should be also supported for passive key expiration for drivers not supporting ttl.

> Refactor R2 with a nested key (id say make the breaking change as its only 4d old, would give full type safety, and we discussed this already)

👍🏼

pi0:
If it is only for type, surely we can add `ttl?` type to generic `TransactionOptions` in this PR 👍🏼

Hebilicious:
@pi0 I would be happy to revisit splitting TransactionOptions once it's actually needed. For now I've implemented everything in this PR according to your requirements.
http and r2 needs to be slightly tweaked because we are no longer using any for TransactionOptions. This doesn't affect the runtime.

pi0:
Thanks for the updates. As an update, i still need to double check all changes there are too many changes some are unrelated to the feature we want to add.

Hebilicious:
> Thanks for the updates. As an update, i still need to double check all changes there are too many changes some are unrelated to the feature we want to add.

@pi0 I had to change these things because according to the docs the put api doesn't support ttl, only the [bulk](https://developers.cloudflare.com/api/operations/workers-kv-namespace-write-multiple-key-value-pairs) API does, and updating the types has a slightly larger blast radius.

pi0:
PR is stalled and with too many changes, have to rework this. 

(but again really appreciate your previous help on this ❤️ )

PR #260, opened 2023-07-04
fix: return `undefined` instead of `null` for non existing keys

### LINKED ISSUE #205: getItem will return `null` for unexisting keys, instead of `undefined`
### Environment

unstorage 1.5.0
nodejs latest

### Reproduction

https://stackblitz.com/edit/node-besgzq?file=index.js

you can see that getItem for a key that doesn't exist will resolve to `null` but should be resolved to `undefined`

then I initialized new Map and .get returned `undefined` as expected

### Describe the bug

`.getItem` should return `undefined` if the key doesn't exist, but it returning `null`

if many drivers implementation there is an OR condition, like `return data.get(key) || null`. this is wrong IMO and should return `undefined` and not `null`.
Also the documentation states clearly that `undefined` should  be returned.

Null might be a misleading return as it might be considered as intentional value and not a key miss.

### Additional context

_No response_

### Logs

_No response_

pi0:
### 🔗 Linked issue

Resolves #205

### ❓ Type of change

### 📚 Description

This is a rework of #206

By changes to this PR, storage returns `undefined` instead of `null` when a key does not exist. 

On the downside, this change makes unstorage less consistent with local storage and Cloudflare KV when `getItem` returns `null` when the key does not exist.

On the benefits, more thinking this is less a breaking change for users but actually a fix since we were already listing `null` as a valid Storage value (which _does_ exist!). And we reduce use of evil null. (https://github.com/sindresorhus/meta/discussions/7)

### 📝 Checklist

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,7 +1,6 @@
 {
   "extends": ["eslint-config-unjs"],
   "rules": {
-    "unicorn/no-null": 0,
     "unicorn/prevent-abbreviations": 0,
     "@typescript-eslint/no-non-null-assertion": 0,
     "unicorn/prefer-string-replace-all": 0
--- a/docs/content/2.usage.md
+++ b/docs/content/2.usage.md
@@ -304,5 +304,5 @@ In [strict mode](https://www.typescriptlang.org/tsconfig#strict), will also retu
 ```ts
 "use strict";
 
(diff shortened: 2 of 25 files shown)


pi0:
PR stalled. Let's track via #205

PR #269, opened 2023-07-08
feat(vercel-kv): support `ttl` for `setItem`

azophy:
here we add new option 'ttl' both for driver setup config & for `set()` method. the interface & implementation took notes from existing redis driver.

### 🔗 Linked issue

this resolve #266

### ❓ Type of change

- [x] 📖 Documentation (updates to the documentation, readme, or JSdoc annotations)
- [ ] 🐞 Bug fix (a non-breaking change that fixes an issue)
- [x] 👌 Enhancement (improving an existing functionality like performance)
- [ ] ✨ New feature (a non-breaking change that adds functionality)
- [ ] 🧹 Chore (updates to the build process or auxiliary tools and libraries)
- [ ] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

### 📚 Description

I have implemented the proposal in #266 . To test it, I have run `pnpm vitest run --coverage` by giving my Vercel KV's credential, and the result shows that it has passed completely. 

image

However because it only uses existing Vercel KV's test suite, it shouldn't have tested the new key expiration functionality. And I'm not sure how I should create the test for this particular feature. I've looked into Redis driver test suite, but I couldn't find section for testing key expiration which I could follow. Maybe other maintainers could give me some pointer? @pi0 @Hebilicious 

### 📝 Checklist

- [x] I have linked an issue or discussion.
- [x] I have updated the documentation accordingly.
- [ ] create appropriate tests

Issue #262, opened 2023-07-05
 standard key expiration for all supported drivers


Hebilicious:
### Describe the feature

Implement a `ttl` top level option for setting a ttl in seconds at the object level for all drivers.

```ts
await setItem(key, item, { ttl: 30 })
```

### Additional information

- Github driver supports global a global ttl option that isn't related to expiration
- Redis driver supports global and setItem ttl options that are related to expiration

For unsupported / partial support drivers, a manual expiration can be implemented by attaching custom unstorage metadata to the storage itself. 

#### Unsupported drivers

- [x] [Azure App configuration](https://stackoverflow.com/questions/73705193/caching-options-for-azure-app-configuration-with-js-sdk)
- [x] [Azure Storage Table](https://stackoverflow.com/questions/9025755/azure-table-storage-expiration)
- [x] fs
- [x] github
- [x] http (altough we could forward ttl to cache-control max age I don't think it makes sense)
- [x] planetscale
- [x] localstorage 
- [x] session-storage
- [x] memory (We could implement with `setTimeout`)

#### Partial support drivers

- [x] R2: This can be set at the [bucket level](https://developers.cloudflare.com/r2/buckets/object-lifecycles/), but not at the object level
- [x] [mongodb](https://www.mongodb.com/docs/manual/tutorial/expire-data/#expire-documents-after-a-specified-number-of-seconds): requires extra indexes

```[tasklist]
### TTL driver native support
- [ ] https://github.com/unjs/unstorage/pull/271
- [x] redis driver
- [ ] https://github.com/unjs/unstorage/issues/266
- [ ] https://github.com/unjs/unstorage/issues/32
- [ ] https://github.com/unjs/unstorage/pull/289
- [ ] Azure Cosmos
- [ ] Azure KV
- [ ] Azure SB
- [ ] lru-cache
```

- [X] Would you be willing to help implement this feature?

YTaoger:
When will this feature be updated?🚀@Hebilicious

itpropro:
I will implement this for the Azure Key Vault and the CosmosDB driver, but Azure Storage Table/Blob and Azure App Config are definitely unsupported natively.
For the general approach, I am not really sure what the benefit is for having the ttl in the metadata treated in another way than normal metadata. I would prefer to just support ttl in a standard way where it is natively supported and leave the rest to custom metadata fields defined by the user.
[...]

PR #271, opened 2023-07-16
refactor: separate set item and get item options types

Hebilicious:
### 🔗 Linked issue

#262 

unblock #255
unblock #269 
 

### ❓ Type of change

### 📚 Description

@pi0 Instead of refactoring all the drivers, I think we can track expiration support in #262. 
This PR just adds support at the type level. 

### 📝 Checklist

### DIFF
--- a/src/drivers/cloudflare-r2-binding.ts
+++ b/src/drivers/cloudflare-r2-binding.ts
@@ -53,12 +53,12 @@ export default defineDriver((opts: CloudflareR2Options) => {
     async setItem(key, value, topts) {
       key = r(key);
       const binding = getBinding(opts.binding);
-      await binding.put(key, value, topts);
+      await binding.put(key, value, topts as any);
     },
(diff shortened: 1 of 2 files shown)


pi0:
For simplification (of typing driver options), i think we can keep using one interface than two and introduce two if needed.

For ttl, i think it actually even makes sense to type possibly supported `ttl` in get operations for two reasons:
- A single options (including possibly credentials, etc) can be used to be passed into all store operations
- `ttl` can be also respected for passive expiration support (drivers not implementing it, lazily invalidate entries on next get operation)

Closing this issue for maintenance but still happy to discuss about introducing two more types if think it is needed 👍🏼

Hebilicious:
@pi0 Even if you do not want to split them now (and I think it's necessary based on the work done in #255), this PR still adds ttl in the `TransactionOptions` interface... Should I just have this change in #255 instead ?

Issue #275, opened 2023-07-23
Support WebDav


shareable-vision:
### Describe the feature

My goal is to provide ergonomic means for individuals who are not extremely tech savvy to publish their own content as simply as possible. WebDav via NextCloud is nice, because it integrates perfectly with clients' desktop, and they do not need to understand git.

I would like to propose that a storage driver be created for WebDav, leveraging the `webdav` package and optionally providing a webhook endpoint to avoid polling.

Are there technical limitations that I am not considering? 
Is anyone interested in supporting me on this?

Thanks for your consideration.

### Additional information

- [X] Would you be willing to help implement this feature?

PR #276, opened 2023-07-23
feat(drivers): webdav driver

shareable-vision:
### 🔗 Linked issue

Issue #275

### ❓ Type of change

### 📚 Description

Webdav driver for unstorage powered by the [`webdav`](https://github.com/perry-mitchell/webdav-client) NPM package.

### 📝 Checklist

### Considerations

The driver currently only implements read-only functions.

### To-do

- [ ] Implement `setItem` and `removeItem`, leveraging `putFileContents` and `deleteFile`

### DIFF
--- a/docs/content/6.drivers/webdav.md
+++ b/docs/content/6.drivers/webdav.md
@@ -4,7 +4,7 @@ navigation.title: Webdav
 
 # Webdav
 
-The driver is powered by the `[webdav](https://www.npmjs.com/package/webdav)` module, but currently only supports read-only use via standard username/password authentication.
+The driver is powered by the [`webdav`](https://www.npmjs.com/package/webdav) module, but currently only supports read-only use via standard username/password authentication.
 
 Fetches all possible keys once and keep it in cache for 10 minutes. Cache only applies to fetching keys.
 
--- a/docs/content/6.drivers/webdav.md
+++ b/docs/content/6.drivers/webdav.md
@@ -1,0 +1,34 @@
+---
+navigation.title: Webdav
+---
+
+# Webdav
+
+The driver is powered by the `[webdav](https://www.npmjs.com/package/webdav)` module, but currently only supports read-only use via standard username/password authentication.
+
+Fetches all possible keys once and keep it in cache for 10 minutes. Cache only applies to fetching keys.
+
+```js
+import { createStorage } from "unstorage";
+import webdavDriver from "unstorage/drivers/webdav";
+
+const storage = createStorage({
+  driver: gitlabDriver({
+    serverURL: "https://nextcloud27.our-servers.de",
+    username: "user",
+    password: "demo123",
+    // pathPrefix: "/remote.php/dav/files/",
+    // directory: "/",
+    // ttl: 600
+  }),
+});
+```
+
+**Options:**
+
+- `serverURL`: Domain of Webdav host.
+- `username`: Webdav username.
+- `password`: Webdav user password.
+- `pathPrefix`: Prefix to content path. Default is `/remote.php/dav/files/`
+- `directory`: Root of content directory. Default is `/`
+- `ttl`: Filenames cache revalidate time. Default is 600 seconds (10 minutes)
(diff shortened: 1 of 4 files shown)


shareable-vision:
Driver (read-only) is tested and working with Nuxt Content.

pi0:
Hi dear @shareable-vision i love the idea of supporting webdav as a driver!

Checking the webdav npm package used, it seems little bit overcomplex. I think we probably can directly implement a webdav client only using native `fetch` API.

An (old) example to inspire: https://github.com/ai-zelenin/webdav-client-fetch/blob/master/ts/index.ts

Do you think you can try to reimplement it with fetch only and no dependencies? 🙏🏼

shareable-vision:
> Do you think you can try to reimplement it with fetch only and no dependencies? 🙏🏼

That's a great idea. I will work on it.

shareable-vision:
WebDav server sends XML data. Although testing with NextCloud, I'm cross-referencing with other authorities and avoiding any namespaced properties, keeping it minimal and hopefully generally compatible.

Native fetch is great . .. but there is not a Node equivalent of browser-native DOMParser.

If we can use [`xml2js`](https://www.npmjs.com/package/xml2js), then I'm almost done re-implementing read-only driver. Installed in an empty repo, `xml2js` clocks in at about 250kb.

Any idea for reliably parsing XML with no dependencies? What do you think?

—I'm really glad you put me onto this!

shareable-vision:
I see now that Unstorage already pulls in the `jsdom` package, which is used by the LocalStorage driver . . I think I may be able to leverage that to parse the XML (provides web-native features, like DOMParser) with no extra dependencies.

pi0:
Hi dear @shareable-vision. Sorry for answering late and thanks for willing to continue working on this. 

Nice point about XML parser lacking. Looking at `webdav` dependencies, they use [fast-xml-parser](https://www.npmjs.com/package/fast-xml-parser) i would recommend to try it. 

Usng bundlephobia.com, fast-xml-parser is 25KB vs JSDom 2.4 MB.

We might find something even smaller but i think that's a good start.

shareable-vision:
Just pushed re-implementation of read-only version of the Webdav driver using 'fast-xml-parser'.

I will work on implementing `setItem()` and `removeItem()` next.

shareable-vision:
I want to work on this a little longer and make some improvements. One very major improvement that I see now will be to use the `Depth` header —currently implementation is recursively re-fetching subdirectories and awaiting a Promise.all. I tried the header with a faulty value ('Infinity') —now I see that a value of 'infinity' (or a high integer) will eliminate the re-fetching and speed up `getKeys()`.

I'm also mulling over the advantages of polling. Will study the other drivers more about this . . 
I think where I'm landing is that it would be good to provide optional caching and polling, which are each disabled with a default value of 0.

pi0:
Thanks for updates and nice work on this @shareable-vision 

I have marked PR as draft in meantime please ping when felt ready

I will also check locally in coming days and try to help if fine for you 👍🏼

shareable-vision:
Perfect.

shareable-vision:
**Ready for review.**

`setItem()` and `removeItem()` are implemented, honoring all WebDAV specs requirements.

`setItem()` will ensure that the target parent directory exists by making MKCOL requests on each non-existent ancestor directory. If 'type' key on opts is set to 'directory', setItem() will make a directory on WebDAV storage instead of uploading a file.

**Notes**
I'm not sure if the PUT requests may be optimized by using a ReadableStream. Probably because they are already utf8-encoded, this won't make any difference . . 

**What should be the desired behavior of the `ttl` key and caching?** In my mind, the most optimal solution for WebDAV is for the consumer of the driver to receive a Webhook from the WebDAV service and then intentionally re-fetch keys/meta. In that case, the consumer would know precisely when re-fetching is required and no cache should be implemented. In such a case, re-implementing the same syncPromise pattern that the Github driver uses wouldn't make sense . . Optimally one would not want to continuously poll or re-fetch content from a slow WebDAV server in which case caching should be totally disabled when using `getMeta`, `getItem`, etc. My thinking before was that an undefined `ttl` might signify ("Do not cache"), however another reasonable interpretation would be (TTL of zero or undefined => Always re-fetch), which would be the opposite of my intention. I need help making a wise decision about this.

shareable-vision:
Driver is ready for review.

PR #281, opened 2023-07-31
refactor: fix typo in `removeMeta` option for `removeItem`

### LINKED ISSUE #280: removeMeta option for removeItem type and default value incorrect in docs
### Environment

N/A

### Reproduction

removeItem("foo:bar", true) will remove meta
removeItem("foo:bar", {removeMeta: true}) won't remove meta
removeItem("foo:bar", {removeMata: true}) will remove meta
removeItem("foo:bar") wont' remove meta

### Describe the bug

removeMeta option is misspelled as "removeMata" meaning setting "removeMeta" option doesn't work

Also docs imply removeMeta is true by default which is not the case leading to unexpected behaviour when not supplying removeMeta option

### Additional context

Also I noticed the removeItem function is commented as going to be removed in the next major release, why is that?

### Logs

_No response_

Timbological:
### 🔗 Linked issue

#280 

### ❓ Type of change

- [x] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

### 📚 Description

Fixed typo of removeMeta option in removeItem function as "removeMata", updated docs to reflect default behaviour or removeItem function when removeMeta argument is not specified

It resolves the problem of the option being spelled incorrectly and therefore the correct spelling not working. It would be a breaking change if someone has realized that it was spelled incorrectly and therefore has been using the incorrect spelling.
It also corrects the docs which imply that not providing a removeMeta argument to removeItem will result in the metadata being deleted which is not the case

Resolves #280 

### 📝 Checklist

- [X] I have updated the documentation accordingly.


pi0 (approved):
Thanks for sharp eyes 💯  (i have added a small backward compatibility check to avoid breaking users)

PR #286, opened 2023-08-14
docs: fix memory driver description

mrbrianevans:
### ❓ Type of change

### 📚 Description

The implementation uses Map rather than Set. 
Aligned docs to implementation.
See https://github.com/unjs/unstorage/blob/main/src/drivers/memory.ts#L6

### 📝 Checklist


pi0 (approved):
Nice spot thanks!

PR #292, opened 2023-08-24
refactor(cloudflare-kv, cloudflare-r2): move `getBindings` to utils and add default `BUCKET` for r2

oritwoen:
### ❓ Type of change

### 📚 Description

The `getBindings` function was used in 2 files. This PR:

- moves the `getBindings` function to `utils`
- adds default `binding` value for `R2`
- adds additional references in the documentation

### 📝 Checklist


src/drivers/utils/index.ts
```diff
@@ -39,3 +40,31 @@ export function createRequiredError(driver: string, name: string | string[]) {
   }
   return createError(driver, `Missing required option \`${name}\`.`);
 }
+
+export function getBinding(binding: KVNamespace | R2Bucket | string) {
```
pi0, line 44: Can you please move this to `utils/cloudflare.ts`? 🙏🏼

oritwoen: Sure, good idea. I have made changes :)

pi0:
(sorry for delay on review. i will do it asap in next days to release 👍🏼 )

oritwoen:
@pi0 

I've been thinking about additional changes lately. There will be more HTTP drivers for Cloudflare in the future. It would be good to put general functions and types in one place:

- https://github.com/unjs/unstorage/blob/main/src/drivers/cloudflare-kv-http.ts#L94
- https://github.com/unjs/unstorage/blob/main/src/drivers/cloudflare-kv-http.ts#L95

etc.

Creating a new PR at this point seems "problematic" because an identical file would be created as in this PR. So maybe change the title of this request and include generalized http functions for cloudflare here as well?

oritwoen:
@pi0 Can you address my comment above?

pi0:
Thanks again! I have made few more refactors to make types working.

PR also welcome to prepare types refactoring we might make `utils/cloudflare.types.ts` to make them more maintainable 💯

PR #306, opened 2023-09-21
refactor(redis): use named import for `Redis`

tobiasdiez:
### ❓ Type of change

### 📚 Description

[Since a few versions](https://github.com/redis/ioredis/blob/ec42c82ceab1957db00c5175dfe37348f1856a93/CHANGELOG.md?plain=1#L35-L42), ioredis supports named imports. This should improve esm compatibility, and maybe at some point enable treeshaking.

### 📝 Checklist

### DIFF
--- a/src/drivers/redis.ts
+++ b/src/drivers/redis.ts
@@ -1,5 +1,6 @@
 import { defineDriver, joinKeys } from "./utils";
-import Redis, {
+import {
+  Redis,
   Cluster,
   ClusterNode,
   ClusterOptions,


pi0:
Thanks for this PR, but do you think it is an essential change? If ioredis keep supporting default export, it can be safer for projects might have an older version of it.

tobiasdiez:
It's not very essential. I have some troubles with the ioredis import (https://github.com/redis/ioredis/issues/1822, mostly when using vitest) and was hoping this is fixed with this change here. But if you think backwards compatibility is an issue, we can still wait with merging this.

pi0:
Can you help on preparing a reproduction for vitest issue?

What else we can do, is to use `import * from` to dedefault in runtime to have both backward and forward (vitest, ESM) compatibility.

tobiasdiez:
It's sadly hard to reproduce, only occurring in certain situations (which seem to involve importing ioredis another time in code that is tested after using unstorage). Will append a reproduction to the issue as soon as I have one.

You are the expert of cjs/esm imports, so I'll let you decide what's the best import here. I got confused by how ioredis is doing its exports
https://unpkg.com/browse/ioredis@5.3.2/built/index.js

pi0:
Thanks again for PR. Since the named export is added somewhere between 5.x, we might make regressions for some older users. I have added a TODO (404c2f1e3e3fea1e517dfbbdcaa2204f3a5c9b28) to make the change for v2.

PR #308, opened 2023-09-22
fix(azure-key-vault-driver): fix character encoding

itpropro:
### ❓ Type of change

### 📚 Description

- Updated key encoding/decoding to avoid decoding errors
- Updated docs with additional usage hints

### DIFF
--- a/docs/content/6.drivers/azure-key-vault.md
+++ b/docs/content/6.drivers/azure-key-vault.md
@@ -5,7 +5,7 @@ Store data in a Azure Key Vault [secrets](https://docs.microsoft.com/en-us/azure
 This driver stores KV information in Azure Key Vault secrets by using the key as secret id and the value as secret content.
 Please be aware that key vault secrets don't have the fastest access time and are not designed for high throughput. You also have to disable purge protection for your key vault to be able to delete secrets. This implementation deletes and purges a secret when it is deleted to avoid conflicts with soft delete.
 
-⚠️ Be aware that this driver stores the keys of your key:value pairs in a encoded way in Key Vault to avoid conflicts with naming requirements for secrets. This means that you will not be able to access manually (outside of unstorage) created secrets inside your Key Vault, as long as they are not encoded in the same way.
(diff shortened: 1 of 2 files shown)


docs/content/6.drivers/azure-key-vault.md
```diff
@@ -5,6 +5,8 @@ Store data in a Azure Key Vault [secrets](https://docs.microsoft.com/en-us/azure
 This driver stores KV information in Azure Key Vault secrets by using the key as secret id and the value as secret content.
 Please be aware that key vault secrets don't have the fastest access time and are not designed for high throughput. You also have to disable purge protection for your key vault to be able to delete secrets. This implementation deletes and purges a secret when it is deleted to avoid conflicts with soft delete.
 
+⚠️ Be aware that this driver stores the keys of your key:value pairs in a encoded way in Key Vault to avoid conflicts with naming requirements for secrets. This means that you will not be able to access manually (outside of unstorage) created secrets inside your Key Vault, as long as they are not encoded in the same way.
```
pi0, line 8:
```suggestion
⚠️ Be aware that this driver stores the keys of your `key:value` pairs in an encoded way in Key Vault to avoid conflicts with naming requirements for secrets. This means that you will not be able to access manually (outside of unstorage) created secrets inside your Key Vault, as long as they are not encoded in the same way.
```

src/drivers/azure-key-vault.ts
```diff
@@ -107,9 +107,9 @@ export default defineDriver((opts: AzureKeyVaultOptions) => {
 });
 
 const base64Map: { [key: string]: string } = {
-  "=": "e",
-  "+": "p",
-  "/": "s",
+  "=": "-e-",
+  "+": "-p-",
+  "/": "-s-",
```
pi0, line 112: Would it makes sense / be possible that we use url encoding? (`=` ~> `%3D`)

itpropro: Unfortunately not, as Key Vault only supports dashes besides alphanumeric characters. This replacement is only necessary because of the potential characters that could result in a base64 encoding. B64 was the only encoding I found that is available without adding an additional dependency and without requiring a custom implementation of an encoding scheme. As dashes are the only allowed characters and there were too many edge cases with replacing the special characters of b64 with alphanumeric characters or only one dash, this was the least overhead solution.
I can add an additional test of course to validate the encoding.

pi0:
Thanks!

PR #319, opened 2023-10-12
feat: add `opfs` driver

### LINKED ISSUE #231: Add Origin Private File System (OPFS) API Support
### Describe the feature

[OPFS](https://webkit.org/blog/12257/the-file-system-access-api-with-origin-private-file-system/) is a new API for storing files using the browser it would cover custom file usecases and enable a less memory intensive file system storage system on the web

### Additional information

loilo:
### 🔗 Linked issue

Resolves #231

### ❓ Type of change

### 📚 Description

This PR adds a driver for the browser-based [origin private file system](https://developer.mozilla.org/en-US/docs/Web/API/File_System_API/Origin_private_file_system), as requested in #231.

Strictly speaking, `opfs` is not an exactly accurate name as the driver actually supports _any_ file system following the [File System API](https://developer.mozilla.org/en-US/docs/Web/API/File_System_API). This PR sticks to the `opfs` naming for two reasons:

1. At the time of writing, there are two browser APIs following the File System API: the [File System Access API](https://wicg.github.io/file-system-access/), which allows accessing working on a user's actual file system, and the origin private file system (OPFS).
   Of these, only the OPFS is supported across browser engines and is therefore the default file system used by this driver.
2. To use a more accurate name for the driver, it should somehow refer to the "File System API". This is an incredibly ambiguous term, so something more clear like the OPFS should be preferred.

The driver and its tests are modeled quite closely after the `fs-lite` driver, especially with regards to the options and the tests.

Some key differences are:

#### The `base` Option

The `base` option is available, yet not required.

Other than with Node's `fs` module, it's safe for the File System API to operate on its root directory since it only exists in a sandboxed area of the device's file system.

While namespacing with `base` should be considered a best practice (which is why it's also used in the documentation example), requiring it can be a footgun, especially when working with alternative file systems.

#### The `fs` Option

As described above, this driver can work with file systems other than the OPFS.
Therefore, there is the additional `fs` option which can be passed an alternative `FileSystemDirectoryHandle` (or a Promise resolving to one).

One way to use this would be to work with the user's real file system in supported browsers.
A naïve way to use that option would look like this:

```js
opfs({ fs: window.showDirectoryPicker() })
```

Of course, this code is silly as it has no checks for browser support or other safe guards, but technically, it works.

#### The `readOnly` and `noClear` Options

The `readOnly` and `noClear` options have been carried over from the `fs`/`fs-lite` drivers, but other than with those, they have also been added to the documentation.

I'm not sure whether they are currently _missing_ from the Node FS driver docs or whether they've been omitted on purpose.

If so, please request a change to drop them from the OPFS documentation as well.

#### Different `hasItem` Behavior

The `hasItem` method returns `false` when the target path exists but is a directory.

To be fair, I consider this a bug in the `fs`/`fs-lite` drivers. Since a directory can not be accessed with `getItem`, it should not report to exist via `hasItem` either.

#### Substitutions for Node Built-ins

Since the driver should run in the browser, it uses naïve reimplementations of some `node:path` functions used in `fs-lite`. They are not as sophisticated as the ones provided by Node, but they should easily be up to the task of handling path-like keys while being a lot leaner in size.

### 📝 Checklist

### DIFF
--- a/src/drivers/opfs.ts
+++ b/src/drivers/opfs.ts
@@ -1,0 +1,109 @@
+import { defineDriver } from "./utils";
+import {
+  DRIVER_NAME,
+  exists,
+  getFileObject,
+  joinPaths,
+  normalizePath,
+  readFile,
+  readdirRecursive,
+  remove,
+  removeChildren,
+  unlink,
+  writeFile,
+} from "./utils/opfs-utils";
+
+export interface OPFSStorageOptions {
+  /**
+   * The filesystem root to use
+   * Defaults to the OPFS root at `navigator.storage.getDirectory()`
+   */
+  fs?: FileSystemDirectoryHandle | Promise<FileSystemDirectoryHandle>;
+
+  /**
+   * The base path to use for all operations
+   * Defaults to the root directory (empty string)
+   */
+  base?: string;
+
+  /**
+   * A callback to ignore certain files in getKeys()
+   */
+  ignore?: (path: string) => boolean;
+
+  /**
+   * Whether to ignore any write operations
+   */
+  readOnly?: boolean;
+
+  /**
+   * Whether to disallow clearing the storage
+   */
+  noClear?: boolean;
+}
+
+export default defineDriver<OPFSStorageOptions | undefined>(
+  (opts: OPFSStorageOptions = {}) => {
+    const fsPromise = Promise.resolve(
+      opts.fs ?? navigator.storage.getDirectory()
+    );
+
+    opts.base = normalizePath(opts.base ?? "");
+
+    const resolve = (path: string) => joinPaths(opts.base!, path);
+
+    return {
+      name: DRIVER_NAME,
+      options: opts,
+      async hasItem(key) {
+        return exists(await fsPromise, resolve(key), "file");
+      },
+      async getItem(key) {
+        if (!(await exists(await fsPromise, resolve(key), "file"))) return null;
+        return readFile(await fsPromise, resolve(key), "utf8");
+      },
+      async getItemRaw(key) {
+        if (!(await exists(await fsPromise, resolve(key), "file"))) return null;
+        return readFile(await fsPromise, resolve(key));
+      },
+      async getMeta(key) {
+        const file = await getFileObject(await fsPromise, resolve(key));
+        if (!file) return null;
+
+        return {
+          mtime: new Date(file.lastModified),
+          size: file.size,
+          type: file.type,
+        };
+      },
+      async setItem(key, value) {
+        if (opts.readOnly) return;
+
+        return writeFile(await fsPromise, resolve(key), value);
+      },
+      async setItemRaw(key, value) {
+        if (opts.readOnly) return;
+
+        return writeFile(await fsPromise, resolve(key), value);
+      },
+      async removeItem(key) {
+        if (opts.readOnly) return;
+
+        return unlink(await fsPromise, resolve(key));
+      },
+      async getKeys() {
+        return readdirRecursive(await fsPromise, resolve(""), opts.ignore);
+      },
+      async clear() {
+        if (opts.readOnly || opts.noClear) return;
+
+        if (opts.base!.length === 0) {
+          // We cannot delete an OPFS root, so we just empty it
+          await removeChildren(await fsPromise, resolve(""));
+        } else {
+          await remove(await fsPromise, resolve(""));
+        }
+      },
+    };
+  }
+);
--- a/src/drivers/utils/opfs-utils.ts
+++ b/src/drivers/utils/opfs-utils.ts
@@ -1,0 +1,394 @@
+import { createError } from ".";
+
+export const DRIVER_NAME = "opfs";
+
+function ignoreNotfoundError(error: any): null {
+  if (error.name === "NotFoundError") return null;
+  throw error;
+}
+
+/**
+ * Normalize a path, removing empty segments and leading/trailing slashes
+ */
+export function normalizePath(path: string): string {
+  const normalizedWrappedPath = `/${path}/`
+    // Replace colons with slashes
+    .replace(/:/g, "/")
+
+    // Remove . segments
+    .replace(/\/\.\//g, "/")
+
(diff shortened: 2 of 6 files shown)


loilo:
An additional thought about naming: I explained my decision to stick with the `opfs` naming for this driver. However, I only considered the browser as an environment.

To this date, I have not had any involvement with any serverless environments, but I am aware that those are increasingly modeled after standardized Web APIs. So, somebody with more exposure to this topic should probably check if any established cloud provider uses the File System API in their services already. In that case, it may be worth reconsidering the naming of this driver (and possibly also drop the OPFS default and instead improve the description of the `fs` option in the docs with examples, so that users searching for "OPFS" also get pointed to the File System API driver).

src/drivers/opfs.ts
```diff
@@ -0,0 +1,109 @@
...
+  /**
+   * Whether to disallow clearing the storage
+   */
+  noClear?: boolean;
+}
+
+export default defineDriver<OPFSStorageOptions | undefined>(
+  (opts: OPFSStorageOptions = {}) => {
+    const fsPromise = Promise.resolve(
+      opts.fs ?? navigator.storage.getDirectory()
```
pi0, line 48: We might have a runtime protection to see if `navigator?.storage` is available and otherwise throw an explicit error.

Also this needs to be via a function like `await getFS()`. Unstorage drivers are expected to be lazy initialized.

src/drivers/opfs.ts
```diff
@@ -0,0 +1,109 @@
...
+
+  /**
+   * Whether to ignore any write operations
+   */
+  readOnly?: boolean;
+
+  /**
+   * Whether to disallow clearing the storage
+   */
+  noClear?: boolean;
```
pi0, line 42: These two options are nice but i introduced them to main `fs` mainly because of security/legacy reasons of unintentional filesystem removals. If not really necessary, i would prefer to not introduce them in first step but later think of a generic unstorage option for readonly mountpoints supporting for all drivers.

src/drivers/utils/opfs-utils.ts
```diff
@@ -0,0 +1,394 @@
...
+export function normalizePath(path: string): string {
+  const normalizedWrappedPath = `/${path}/`
+    // Replace colons with slashes
+    .replace(/:/g, "/")
+
+    // Remove . segments
+    .replace(/\/\.\//g, "/")
+
+    // Remove duplicate slashes
+    .replace(/\/{2,}/g, "/");
```
pi0, line 22: Perf: Can we merge these into one regex?

src/drivers/utils/opfs-utils.ts
```diff
@@ -0,0 +1,394 @@
...
+  if (segments.length === 0) return root;
+
+  const create = Boolean(opts.create);
+  const handleType: UnspecificHandleType = opts.type ?? "any";
+
+  // If possibly creating the handle, a specific type must be provided
+  if (create && handleType !== "file" && handleType !== "directory") {
+    throw createError(
+      DRIVER_NAME,
+      'Invalid handle type, must be "file" or "directory" when creating'
```
pi0, line 152: Tip: Using backticks, makes better console formatting when consola is enabled. Feel free to update other messages if i missed.

```suggestion
      'Invalid handle type, must be `file` or `directory` when creating'
```

pi0 (commented):
Added some comments but really nice work ❤️ 💯

src/drivers/utils/opfs-utils.ts
loilo: I generally prefer code clarity over micro optimizations, but I'm also not too well-versed regarding performance impact of regexes, so you're the boss of course. 😁 
Reduced normalization to one single regex.

src/drivers/utils/opfs-utils.ts
loilo: Seems like this is the only occurrence, resolved.

src/drivers/opfs.ts
loilo: Removed `noClear` and `readOnly` options. A generic options sounds like a good idea for this.

src/drivers/opfs.ts
loilo: Alright, done.

loilo:
Hey all. Is there anything I can help with to finish this one? :)

ManUtopiK:
Hi!
Thanks a lot for your driver!
It works very well, but I have an issue while keeping the handle after a page refresh.

First, I want to save the name of the directory chosen by the user. The only way to do that is to manage the `showDirectoryPicker()` outside the driver and pass the instance to `fs`. That works.
Now, I want to mount the opfs driver just after a page refresh. Normally, we can do that as long as the page session is active (like the sessionStorage).
Here is a part of my code :
```js
async function requestNativeBrowserFSPermission(dirHandle: FileSystemDirectoryHandle) {
    const opts: any = {}
    opts.writable = true
    // For Chrome 86 and later...
    opts.mode = 'readwrite'
    const perms = await dirHandle.requestPermission(opts)
  
    return perms === 'granted'
}

let fs
if (source.base === 'directoryPicker') {
    fs = await window.showDirectoryPicker()
    if (!(await requestNativeBrowserFSPermission(fs))) return 
} else {
    const dir = await navigator.storage.getDirectory()
    fs = await dir.getDirectoryHandle(source.base, { create: true })
  
    if (!(await requestNativeBrowserFSPermission(fs))) return
}

storage.mount(`opfs:${fs.name}`, opfsDriver({ fs }))
return fs.name
```
The first time, `source.base = 'directoryPicker'`, so it mounts the driver and ask to pick a directory.
After chosen a folder, I have to accept one browser popup to read files, and one more popup throw to ask for the permission to write files.
It will be nice to ask one time directly to `readwrite`.
So, after accepted the two browser popups, I can see the files. But if I refresh the page, the `source.base` is the name of the previous folder I've chosen (I keep it in the sessionStorage). It doesn't ask for permissions, and I can see that the driver is mounted, but nothing appear :angry: 
There's no error in the console.

I don't know where I'm wrong. Have you got an idea on this ?
Thx!

PR #338, opened 2023-11-15
refactor(drivers): combine `localStorage` and `sessionStorage`

### LINKED ISSUE #203: replace Session Storage and Local Storage drivers with Web Storage driver
### Describe the feature

session and locale storage is really same kind of storage so it does not make sense to have two separate drivers for them 

see 
- https://developer.mozilla.org/en-US/docs/Web/API/Storage
- https://developer.mozilla.org/en-US/docs/Web/API/Web_Storage_API

it's also one interface in typescript for both objects

- https://github.com/microsoft/TypeScript/blob/main/src/lib/dom.generated.d.ts#L21520
- https://github.com/microsoft/TypeScript/blob/main/src/lib/dom.generated.d.ts#L25709
- https://github.com/microsoft/TypeScript/blob/main/src/lib/dom.generated.d.ts#L25758

### Additional information

thecotne:
fixes #203

### DIFF
--- a/docs/content/6.drivers/web-storage.md
+++ b/docs/content/6.drivers/web-storage.md
@@ -1,0 +1,26 @@
+---
+navigation.title: Web Storage
+---
+
+# Web Storage
+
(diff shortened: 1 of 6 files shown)


pi0:
Hi dear @thecotne. it took long to be back on this and some changes stalled. I have made a rework in #530 but thanks again for time on PR and your nice idea for refactor ❤️

Issue #341, opened 2023-11-16
Make watch callback more flexible?


dvlden:
### Describe the feature

I find it weird that `watch` method doesn't actually provide me `k:v` pair for example.
Is there any specific reasoning why `watcher` strictly returns only `event` and `key` and can't be customized further?

Building a browser storage driver is not quite good this way. I'd expect in watcher to receive a key and value on update and only key on delete, for example. 

With current implementation, I'd need to `get` the value from the storage, using the `key` it provides me with, which seem like overhead?

### Additional information

- [ ] Would you be willing to help implement this feature?

pwFoo:
Hello,
played arround with unstorage and it should take over and replace some of my features code parts out of the box! 
Thanks for that awesome project 🥇 👍 

`watch` could much more powerful with callback arguments changed to `(key, value, oldValue, event) => {}`. 

1. **value:** Because value should be available during save data and trigger event a `get` overhead can be avoided?
2. **oldValue:** Would be cool for some use cases and helps to have more precise `event` type (added, updated or deleted) and more flexibility?
3. **event filtering:** Ok, just optional if easy to implement because it could be done by a wrapper arround the watch callback... filtering events by key match or regexp / wildcard based filter functions to listen to specific changes

I would love to see value and oldValue for my needs :)

PR #342, opened 2023-11-16
feat(storage): allow record in watch callback

dvlden:
### 🔗 Linked issue

#341 

### ❓ Type of change

### 📚 Description

It provides flexibility in what the watch callback can return, when changes are detected.
Sometimes it can be highly useful to get both the `key` and the `value` of the property that was changed. I guess sometimes it can be used for some additional properties too.

### 📝 Checklist

> I did not find in the docs any mentions of what callback must be defined as.

### DIFF
--- a/src/storage.ts
+++ b/src/storage.ts
@@ -69,13 +69,15 @@ export function createStorage<T extends StorageValue>(
       }));
   };
 
-  const onChange: WatchCallback = (event, key) => {
+  const onChange: WatchCallback = (event, resource) => {
     if (!context.watching) {
       return;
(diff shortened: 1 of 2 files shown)


pi0:
Hi. I appreciate your work on this PR but not all storage driver watchers get the value on watch callback not this PR adds support for any driver. 

We can do it by introducing a third optional `value?: unknown` parameter to the callback AND consider which drivers already have it in their watch event (I guess currently only web-based drivers) 

Closing PR but feel free to open a new with above proposal.

PR #343, opened 2023-11-20
docs: use nuxt ui pro

Flosciante:


pi0:
Moving to https://github.com/unjs/unstorage/pull/374

PR #353, opened 2023-12-10
feat(driver): add export and info about `dragonfly`

### LINKED ISSUE #294: Dragonfly driver (Redis)
### Describe the feature

Dragonfly is referred to as the modern version of Redis and The Fastest In-Memory Data Store.

> In benchmarks, Dragonfly showed a 25X increase in throughput compared to Redis, crossing 3.8M QPS on c6gn.16xlarge.

- https://www.dragonflydb.io/

### Additional information

oritwoen:
### 🔗 Linked issue

Close #294 

### ❓ Type of change

### 📚 Description

This PR adds information about dragonfly in the documentation and an additional export for better DX.

### 📝 Checklist

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -12,6 +12,7 @@ export const builtinDrivers = {
   cloudflareKVBinding: "unstorage/drivers/cloudflare-kv-binding",
   cloudflareKVHTTP: "unstorage/drivers/cloudflare-kv-http",
   cloudflareR2Binding: "unstorage/drivers/cloudflare-r2-binding",
+  dragonfly: "unstorage/drivers/redis",
   fs: "unstorage/drivers/fs",
   fsLite: "unstorage/drivers/fs-lite",
   github: "unstorage/drivers/github",
--- a/docs/content/6.drivers/dragonfly.md
+++ b/docs/content/6.drivers/dragonfly.md
@@ -1,0 +1,26 @@
+# Dragonfly
+
+Store data in a [Dragonfly](https://www.dragonflydb.io/) storage using [ioredis](https://github.com/luin/ioredis).
+
+::alert{type="warning"}
+Currently, Dragonfly uses the same API and driver as [redis](/drivers/redis).
+::
+
+```js
+import { createStorage } from "unstorage";
+import dragonflyDriver from "unstorage/drivers/dragonfly";
+
+const storage = createStorage({
+  driver: dragonflyDriver({
+    base: "unstorage",
+    host: 'HOSTNAME',
+    tls: true as any,
+    port: 6380,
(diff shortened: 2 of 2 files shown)


src/index.ts
```diff
@@ -12,6 +12,7 @@ export const builtinDrivers = {
   cloudflareKVBinding: "unstorage/drivers/cloudflare-kv-binding",
   cloudflareKVHTTP: "unstorage/drivers/cloudflare-kv-http",
   cloudflareR2Binding: "unstorage/drivers/cloudflare-r2-binding",
+  dragonfly: "unstorage/drivers/redis",
```
pi0, line 15: We might need a entry file even if it is simple reexport

At least defaults such as `tls` and `port` could be enabled by default also this way (haven't tried, if needed)

oritwoen: @pi0 Your own export is a good idea.

However, in my opinion, first the Redis interface in the repository needs to be better organized. There are more and more libraries that have a virtually identical interface and unstorage will certainly support more of them in the future.

The difference between `Redis`, `Dragonfly`, `VercelKV`, `Upstash` etc. is nominal and it is possible to create a universal interface for all drivers that use redis.

With dragonfly `unstorage` would already have 3 drivers using redis, so it's high time to organize it better.

I have an idea to create a separate PR in which the current drivers (Redis and VercelKV) would be separated into `utils` and thus, for each new driver that uses redis, it would be possible to import ready-made functions with their own configurations instead of constantly repeating the code.

And then go back to this PR, do a separate export and use these utils. What do you think about it?

pi0: We can extend from redis and refactor other drivers later if can save us maintenance time but give same possibilities why not 💯 

However keep in mind, while all these are redis-compatible drivers, some like vercel one, depends on specific SDK like `@vercel/kv` that essentially wraps with default environment variables. Yes, we can remove this layer but it requires us to keep in sync with what env variables vercel provides.

drewbitt:
Are there any remaining changes requested?

src/index.ts
oritwoen: @pi0 I currently don't have a good idea or time on how to deal with the topic mentioned in the comments above.

So I made a clone of the redis driver for dragonfly and corrected and updated the pull request.

Close or merge whatever you think is right. Due to these comments, if something comes to my mind, I will open a discussion or a new pull request.

pi0:
Hi dear @oritwoen sorry for all delay on coming back to this. 

I really appreciate your time on this PR. I had been trying to figure out if dragonfly (client side) needs a specific SDK or any convention worth to have a dedicated driver for but couldn't figure out any reason.

Therefore closing PR for now in favor of [redis driver](https://unstorage.unjs.io/drivers/redis)

Issue #24, opened 2021-09-22
Encrypted storage


pi0:
Either a proxy similar to [prefixStorage](https://github.com/unjs/unstorage#prefixstoragestorage-data-base) or an easy interceptor (#23) that encrypts data and decrypts when reading.

This can be so useful when we do not trust 

Questions:
- Allow encrypt subpaths?
- Allow encrypting per read/write?
- How to use blobs in raw storage? (#21)
- What algorithm use that is both lightweight in pure js AND can be accelerated by runtime?

PR #363, opened 2024-01-02
feat: e2e encryption using `encryptedStorage`

itpropro:
### 🔗 Linked issue

#24 

### ❓ Type of change

- [X] ✨ New feature (a non-breaking change that adds functionality)

### 📚 Description

Implements a new composable function `encryptedStorage` similar to `prefixStorage` that wraps a unstorage instance and offers encryption for values and optionally for keys as well.

- Adds new composable `encryptedStorage` that can also be used in combination with others like `prefixStorage`. Usage is `encryptedStorage(createStorage({ driver }), encryptionKey, true)` where the last parameter (set to `true`) defines if keys should also be encrypted
- Uses `@noble/ciphers` for encryption, as it is 0 deps and works in Node, Bun, Deno and Workers (tested on CF). It supports AES-GCM-SIV (not implemented in Web/Node Crypto) which is nonce misuse resistant, what we need for deterministic key encryption with the same IV/nonce for keys.
- For content encryption, XChaCha20-Poly1305 (which is also used in TLS 1.3) is used. It is a fast modern and future proof cipher and safe to use with random nonces. Nonces are generated with `getRandomValues` from `uncrypto`. The content is wrapped into a `StorageValueEnvelope`:

```ts
export interface StorageValueEnvelope {
  nonce: string;
  encryptedValue: string;
}
```

- The implementation was tested against common drivers like `fs`, `memory`, `redis`, 'lrs' and `azure-blob-storage` for content as well as key encryption and in combination with `prefixStorage`:

```ts
    const storage = createStorage();
    const pStorage = prefixStorage(storage, "foo");
    const encStorage = encryptedStorage(pStorage, encryptionKey, true);
```

- Tests include an encryption example for `fs` driver, an update to the `testDriver` utility to optionally test for content and key encryption as well es tests for the storage server and the combination with `prefixStorage`

### 📝 Checklist

## Summary by CodeRabbit

* **New Features**
  * Added encryptedStorage for transparent end-to-end encryption of values, optional key encryption, and secret rotation support.

* **Documentation**
  * Added an encryption guide describing usage, envelopes, key-prefix behavior, raw workflows, and rotation.

* **Tests**
  * Added comprehensive tests for encryption, raw byte APIs, key encryption, rotation, large values, and prefixed storage.

* **Chores**
  * Build now minifies library chunks at build time for smaller output.


itpropro:
Would appreciate your feedback @pi0, @danielroe, @Atinux. 
If the implementation is fine for you, I would add an option to provide the encryption key via. env variable and write the docs.

test/drivers/netlify-blobs.test.ts
```diff
@@ -5,7 +5,7 @@ import { BlobsServer } from "@netlify/blobs";
 import { resolve } from "path";
 import { rm, mkdir } from "node:fs/promises";
 
-describe("drivers: netlify-blobs", async () => {
+describe.skip("drivers: netlify-blobs", async () => {
```
danielroe, line 8:
```suggestion
describe("drivers: netlify-blobs", async () => {
```

itpropro: That driver’s tests are extremely flaky, which is why I disabled them and forgot about it, thanks for the find :)
What do you think about the implementation @pi0 ?

itpropro:
Anything I can do to bring this forward @pi0 ?

itpropro:
> 🧹 Nitpick comments (1)
> 
> > test/encryption.test.ts (1)> `6-6`: **Prefer generated test vectors over repeated key-like literals.**
> > > These constants are test-only, but they repeatedly trigger secret scanners and are duplicated across cases. Consider deriving deterministic keys in-test to reduce noise and drift.
> > > 
> > > Refactor sketch
> > > ```diff
> > > -  const sivNonce = "ThtnxLK9eCF4OLMg";
> > > -  const secret = "e9iF+8pS8qAjnj7B1+ZwdzWQ+KXNJGUPW3HdDuMJPgI=";
> > > +  const b64 = (s: string) => globalThis.btoa(s);
> > > +  const sivNonce = b64("nonce-12-bytes"); // 12 bytes -> 16-char base64
> > > +  const secret = b64("0".repeat(32)); // 32 bytes
> > > +  const rotatedSecret = b64("1".repeat(32)); // 32 bytes
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > Then reuse `rotatedSecret` instead of redefining `newSecret` per test.
> > > 
> > > Also applies to: 83-83, 104-104, 117-117, 139-139, 171-171
> > > 
> > > 🤖 Prompt for AI Agents
> > > ```
> > > Verify each finding against the current code and only fix it if needed.
> > > 
> > > In `@test/encryption.test.ts` at line 6, Replace the hardcoded base64 key literals
> > > used in tests (e.g., the constant named secret and repeated newSecret instances)
> > > with a single deterministically derived test vector generated at runtime (for
> > > example, derive rotatedSecret/newSecret from a single test-only seed using a KDF
> > > or HMAC-SHA256 so outputs are stable across runs); update all tests to reuse
> > > that derived variable (rotatedSecret) instead of redefining newSecret in each
> > > case so you eliminate repeated key-like literals and secret-scanner noise while
> > > keeping deterministic, reusable test keys.
> > > ```
> 
> 🤖 Prompt for all review comments with AI agents
> 
> ...

Want to have this fixed as well @pi0 or are we good?

Issue #142, opened 2023-02-06
Raw format support tracker


pi0:
Initial issue: #21. Tracking experimental `getItemRaw/setItemRaw` support. Please comment any particular issues and ideas with the current implementation.

- Default serializer is only supported in environments supporting Buffer. (Node.js and Nitro/Nuxt server in any environment)
- Cannot combine string and raw operations on the same key
  - Default serializer only operates on `base64:` strings without json support 
  - This makes issues with `overlay` and `http` backends when translating string operations
- Types are wide to `any` since we don't enforce any return value from drivers implementation.

ashutosh887:
@pi0 Does this need some refractor?

0xAsimetriq:
Was attempting to store `CryptoKey` type, but got an error as it didn't satisfy allowed types (believe some of the listed ones were `ArrayBuffer`, `UTF...`)

dnl13:
Hello,

Since this topic is also flagged as a discussion, I have a short question. My goal: I would like to cache/store WASM models and binaries in indexedDB. Since these models are intended for client-side use only and my NuxtJS Composable is also only called on the client-side, I have some questions about `setItemRaw`.

I have compared both `setItemRaw()` and `set()` from `idb-keyval`. 
With `set()`, I can easily store my Uint8Array in indexedDB on the client-side without any issues.

However, I get the error `"ReferenceError: Buffer is not defined"` due to the serializer in `setItemRaw`, which is understandable because Buffer is not available on the client-side.

My question is whether we can simply add `set` and `get` from indexedDB to our createStorage with the indexedDB-Driver to support these functions on the client-side, or why serialization is necessary at all (shouldn't developers be responsible for serializing data if it can't be stored in indexedDB in the desired format)? The term `setItemRaw` implies that the data is appended without any additional processing, so perhaps the wording isn't ideal.

I am relatively new to this topic and unfortunately do not have very deep knowledge about it. I just wanted to share my observation and ask how I could store my Uint8Array with unstorage and the indexeddb driver without having to convert it to base64, which can be quite performance-intensive for a model of around > 350MB in size.

Attached is a short NuxtJS composable that I tried:

```
import { createStorage } from "unstorage";
import indexedDbDriver from "unstorage/drivers/indexedb";
import { get, set , createStore  } from 'idb-keyval';

export const useModelStore = async () => {

  const store = createStorage({
    driver: indexedDbDriver({ base: "app", dbName: 'model-store', storeName: 'model' }),
  })
[...]

PR #364, opened 2024-01-02
feat: support raw conversion in envs without `Buffer`

itpropro:
### 🔗 Linked issue

#142

### ❓ Type of change

- [X] 👌 Enhancement (improving an existing functionality like performance)

### 📚 Description

Replaced Buffer with native implementation to also run on workers and other environments. It's also faster than Buffer, but in the context of unstorage, the improvement is not significant.

### 📝 Checklist


itpropro:
The netlify tests seem to be very flaky and slow, we should probably disable them (95% of them ran into timeout on my machine) until the driver is fixed or a faster way of testing is found.

GalacticHypernova:
How would this translate to getItemRaw with the web streams? Would it still be possible for the response of the image to be rendered correctly?

src/_utils.ts
```diff
@@ -72,6 +65,34 @@ export function deserializeRaw(value: any) {
     // Return unknown strings as-is
     return value;
   }
-  checkBufferSupport();
-  return Buffer.from(value.slice(BASE64_PREFIX.length), "base64");
+  return genBytesFromBase64(value.slice(BASE64_PREFIX.length));
+}
+
+// Base64 utilities - Waiting for https://github.com/unjs/knitwork/pull/83 // TODO: Replace with knitwork imports as soon as PR is merged
```
manniL, line 71: Can now be used via https://github.com/unjs/undio/pull/8

pi0: Yes indeed. The feature is not releaed yet and I ned to complete node {Buffer} support but we shall adopt undio.

src/drivers/http.ts
```diff
@@ -33,14 +33,14 @@ export default defineDriver((opts: HTTPOptions) => {
...
-      const value = await _fetch(r(key), {
+      const value = await _fetch.native(r(key), {
         headers: {
           accept: "application/octet-stream",
           ...opts.headers,
           ...topts.headers,
         },
       });
-      return value;
+      return new Uint8Array(await value.arrayBuffer());
```
pi0, line 43: This is nice change but has to be split

src/_utils.ts
pi0: will add a TODO to refactor once undio stable

pi0 (approved):
Thanks ❤️

src/drivers/http.ts
pi0: ~> #527

PR #365, opened 2024-01-03
test(http):  add tests for `null` value

itpropro:
### 🔗 Linked issue

closes #268

### ❓ Type of change

- [ ] 📖 Documentation (updates to the documentation, readme, or JSdoc annotations)
- [ ] 🐞 Bug fix (a non-breaking change that fixes an issue)
- [ ] 👌 Enhancement (improving an existing functionality like performance)
- [ ] ✨ New feature (a non-breaking change that adds functionality)
- [X] 🧹 Chore (updates to the build process or auxiliary tools and libraries)
- [ ] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

### 📚 Description

Added null tests to verify correct handling of null values in http driver.

### 📝 Checklist

- [X] I have linked an issue or discussion.
- [ ] I have updated the documentation accordingly.

PR #367, opened 2024-01-06
fix(http, server): handle missing resources with http 404

Aareksio:
### 🔗 Linked issue

https://github.com/unjs/ipx/pull/204

Screenshot 2024-01-06 at 13 31 47

### ❓ Type of change

### 📚 Description

HTTP driver did not properly return `null` when resource was not found. The error was masked by `createStorageServer` newer throwing 404 as actual HTTP server should. This PR makes the `createStorageServer` return 404 and `http` driver to properly catch 404 and return `null`.

I have ran the changes against https://github.com/unjs/unstorage/pull/365 and the behavior stays the same.

### 📝 Checklist


pi0 (approved):
Thanks!

Small behavior changes to be documented in release notes:

- HTTP server throws 404 for uneset values. This requires unstorage clients to update to the latest version to support handling
- HTTP driver throws error for `hasItem` when response is not 404 (previously always returning `null`)

src/drivers/http.ts
```diff
@@ -24,37 +32,23 @@ export default defineDriver((opts: HTTPOptions) => {
...
-        if (err?.response?.status === 404) {
-          return null;
-        }
-
-        throw err;
-      }
+      const value = await _fetch(r(key), {
+        headers: { ...opts.headers, ...tops.headers },
+      }).catch(catchFetchError);
+      return value;
```
Aareksio: Is there a reason you prefer to assign a variable instead of simply returning?

pi0: `return await` could also do it similarly (it is for making stack traces better in some runtimes)

`value` + `return` is mostly conventional across drivers to make debugging easier

PR #370, opened 2024-01-09
feat: add `neon` driver

siegerts:
### ❓ Type of change

### 📚 Description

This PR adds support for Neon Postgres using the Neon Serverless driver.

### 📝 Checklist


src/drivers/neon.ts
```diff
@@ -0,0 +1,99 @@
+import { createError, createRequiredError, defineDriver } from "./utils";
+import { neon, NeonQueryFunction, neonConfig } from "@neondatabase/serverless";
+
+export interface NeonDriverOptions {
```
evanshortiss, line 4: Could this object accept a `driverConfig` of type `NeonConfig` from the `@neondatabase/serverless`. This would allow users to supply other configuration options for the driver, or is that a little too leaky?

evanshortiss: Actually, nevermind. Users could just import `@neondatabase/serverless` and directly modify the `neonConfig`, so does the `fetchConnectionCache ` config need to be handled by this at all?

src/drivers/neon.ts
```diff
@@ -0,0 +1,99 @@
...
+
+export default defineDriver((opts: NeonDriverOptions = {}) => {
+  opts.table = opts.table || "unstorage";
+  neonConfig.fetchConnectionCache = opts.fetchConnectionCache || false;
+
+  let _connection: NeonQueryFunction<false, false>;
+
+  const getConnection = () => {
+    if (!_connection) {
+      if (!opts.url) {
```
evanshortiss, line 33: Would it make more sense to check this immediately, e.g on line 26 above? Fail fast(er)

evanshortiss (commented):
Awesome. I just added two minor questions.

src/drivers/neon.ts
siegerts: Left it in to be consistent with the use in blog reference and docs. Also, added a short section highlight the additional advanced config, if needed.

pi0:
Would be nice if we do have DB0 storage first because this way when unstorage is used in Nitro/Nuxt, we can leverage auto configuration. (There is even a [placeholder page](https://db0.unjs.io/connectors/neon) already!)

Next we can integrate neon with unstorage using generic SQL driver (https://github.com/unjs/unstorage/issues/400) but I see value of db0 provider would be much much more than focusing on unstorage first.

PR #383, opened 2024-01-21
feat: support for `findItems` and `findItemsRaw`

GalacticHypernova:
### ❓ Type of change

### 📚 Description

Currently, the only way to get items in batch is through `getItems`, which is a shorthand for a multi-line `getItem`. It would be nice to have a way to search for items that may match, as we could use it for easier and more convenient batch lookups. For example, when we want to get all the webp images when using the fs driver, we could use:
```ts
const storage = useStorage('fs')
const images = await storage.findItems(".webp")
```

### 📝 Checklist


GalacticHypernova:
Converting to a draft to take the time to work on it. For drivers that do not inherently support enumeration we can simply return an empty array with a warning that the driver doesn't support it.

GalacticHypernova:
I do wonder, how would we make this work for raw getters as well? Should this default to non-raw?

pi0:
Hi. Thanks for the PR but it is something that easily can be done in user-land (your projects) and nothing can benefit from driver natives that is common enough adding it to the unstorage core, makes the library base-size bigger which I tend to avoid.

I think you can refactor your work into a new `fineItems()` exported method tough, please feel free to open a new PR that or an issue to keep tracking.

GalacticHypernova:
> I think you can refactor your work into a new `fineItems()`

I thought about originally changing the findItems method but I didn't want to change something unexpectedly and modifying core functionality. Thanks for the feedback!

pi0:
I think since is for utility, we can make it like `findItems(storage, prefix, { raw: boolean }` wdyt?

Issue #389, opened 2024-01-25
uploadthing driver


pi0:
(context: found about uploadthing from https://github.com/unjs/h3/issues/628)

- https://uploadthing.com
- https://github.com/pingdotgg/uploadthing

Let's make uploads for *developers better ;) 

/cc @juliusmarminge @t3dotgg @markflorkowski let me know if likes to work to make it happen ❤️

juliusmarminge:
We have Nuxt and Vue support coming but are currently blocked by a type bug in Vue https://github.com/vuejs/core/issues/10167

pi0:
That's nice for Vue/Nuxt!! FYI, unstorage scope is a little bit different. It is more a backend layer that allows storing data in agnostic backends (like a filesystem) which can be retrieved to be for example served by another handler and not tied to Vue. Feel free to drop me a DM in X or discord (_pi0, pi0) if like to have a chat!

pi0:
Experimental driver landed in #390

PR #390, opened 2024-01-25
feat: add uploadthing driver

juliusmarminge:
### 🔗 Linked issue

#389 

### ❓ Type of change

### 📝 Checklist


src/drivers/uploadthing.ts
```diff
@@ -0,0 +1,90 @@
...
+  const getClient = () => {
+    return (client ??= new UTApi({
+      apiKey: opts.apiKey,
+      fetch: ofetch.native,
+    }));
+  };
+
+  const getUTFetch = () => {
+    return (utFetch ??= ofetch.create({
+      baseURL: "https://uploadthing.com/api",
```
pi0, line 22: Would it makes sense to support an optional option for baseURL? (mainly to allow unit testing)

src/drivers/uploadthing.ts
```diff
@@ -0,0 +1,90 @@
...
+
+  return {
+    hasItem(key) {
+      // This is the best endpoint we got currently...
+      return getUTFetch()("/getFileUrl", {
+        body: { fileKeys: [key] },
+      }).then((res) => res.ok);
+    },
+    getItem(key) {
+      return ofetch(`https://utfs.io/f/${key}`);
```
pi0, line 37: Maybe, somehow make it configurable too? (not sure about `/api` and this API difference but just a taught to again make testing easier)

src/drivers/uploadthing.ts
```diff
@@ -0,0 +1,90 @@
...
+    },
+    getItemRaw(key) {
+      return ofetch
+        .native(`https://utfs.io/f/${key}`)
+        .then((res) => res.arrayBuffer());
+    },
+    getItems(items) {
+      return Promise.all(
+        items.map((item) =>
+          ofetch(`https://utfs.io/f/${item.key}`).then((res) => ({
```
pi0, line 47: If there is no bulk support in API, you can safely remove `getItems`. unstorage (core) will make same parallel call anyway.

src/drivers/uploadthing.ts
```diff
@@ -0,0 +1,90 @@
...
+            key: item.key,
+            value: res,
+          }))
+        )
+      );
+    },
+    getKeys() {
+      return getClient()
+        .listFiles({})
```
pi0, line 56: Just checking, if `listFiles` supports a base, `getKeys(base)` accepts it (and it can make usage faster specially when there are too many files and one needs to filter)

src/drivers/uploadthing.ts
```diff
@@ -0,0 +1,90 @@
...
+    },
+    removeItem(key, opts) {
+      return getClient()
+        .deleteFiles([key])
+        .then(() => {});
+    },
+    async clear() {
+      const client = getClient();
+      const keys = await client.listFiles({}).then((r) => r.map((f) => f.key));
+      return client.deleteFiles(keys).then(() => {});
```
pi0, line 84: `clear(base)` is suported too. . I guess we might need to support it (client-side or server side) otherwise one call drops entire collection of uploads 🗡️

src/drivers/uploadthing.ts
juliusmarminge: uploadthing currently doesn't support directories, the only filter options this `listFiles` takes are pagination-related

src/drivers/uploadthing.ts
juliusmarminge: utfs is our cloudflare rewriter that handles caching and mapping keys to different regions.

uploadthing.com/api is where everything else goes.

can make it configurable for testing though!

src/drivers/uploadthing.ts
pi0: Oh okay i see. BTW unstorage doesn't too. We just assume `:` and `/` as fs-like seperators for prefixing. Would that be something you consider supporting at some point? (I imagine for quering large amount of files could come in use)

pi0:
We have unit tests you can run with pnpm run vitest (unstoage also targets server). We mainly need to mock endpoints (responding from mobile if couldn’t find it please tell me to push commit)

src/drivers/uploadthing.ts
markflorkowski: Definitely something we would consider supporting

juliusmarminge:
I just realized that in order to support this we'd need to allow setting the key manually. right now uploadthing generates the key for you

juliusmarminge:
> I just realized that in order to support this we'd need to allow setting the key manually. right now uploadthing generates the key for you

@pi0 is there a concept of mapping keys inside unstorage? For example when we setItem we get the key from UT, that we could then map to the unstorage key, i.e:

```ts
storage.setItem("foo")
// internally map foo -> KEY_FROM_UT
storage.getItem("foo")
// key = map["foo"], fetch(utfs.io/f/key)
```

pi0:
I think we mainly need a storage for keys which is usually same storage layer. 

If your API is designed to always return random names, I am up to consider a new change that `setItem` returns the random name. It is a non breaking change and i hope to unblock this presets also any other preset that need this.

(in the meantime, i guess it would be also nice if you consider supporting custom names. let's see which comes first haha)

markflorkowski:
> (in the meantime, I guess it would be also nice if you consider supporting custom names. let's see which comes first haha)

The reason we don't is because names must be URL-safe, and have a length limit.

pi0:
Hmm, I see. What if we make the driver make sure that input is url-safe and respects the limit like any fs does (so it throws an error if it doesn't comply) I think users of the driver, would understandably follow this and probably can benefit more from such control vs random ids. Thoughts?

juliusmarminge:
So I did a thing to track internal vs uploadthing keys - not sure if it aligns with what this project aims to do though - I'll let you be the judge.

Currently passing all tests except `clear` since our API throws a 500 when you try to delete 0 files currently (will fix that momentarily)

markflorkowski:
> Currently passing all tests except clear since our API throws a 500 when you try to delete 0 files currently (will fix that momentarily)

Merged fix for this on the API side

src/drivers/uploadthing.ts
```diff
@@ -4,87 +4,111 @@ import { UTApi } from "uploadthing/server";
...
-  /**
-   * Primarily used for testing
-   * @default "https://uploadthing.com/api"
-   */
-  uploadthignApiUrl?: string;
-  /**
-   * Primarily used for testing
-   * @default "https://utfs.io/f"
-   */
-  uploadthingFileUrl?: string;
```
juliusmarminge: these makes no sense as we don't allow this override to the `UTApi` anyways

pi0: This option provides two benefits:

- Unit testing the driver internally without actually calling UTAPi but mocking it in unstorage end-to-end tests
- Allow users a way to Proxy your API if their server deployment requires this for any reason

I think both are internal and advanced cases. But if you really want to discourage users more to change them, we can definitely make it better with naming and JSDocs.

juliusmarminge:
current approach only works for a long-running server since it depends on that in-memory map. we're thinking about how to support user-provided keys

src/drivers/uploadthing.ts
juliusmarminge: Yes but that would only override the URL for the fetch calls that doesnt use utapi. All fetches made within utapi will still use uploadthing.com (although we have an internal env variable that overrides it)

juliusmarminge:
struggling getting msw to intercept the requests 🤔 

CleanShot 2024-01-25 at 23 15 52

pi0:
yeah. msw is not the best in class nor supports ofetch which I am aware of and also the fix but still hesitated because by doing it we add an overhead to every user of it only because of msw. 

I will try to check it ASAP.

juliusmarminge:
@pi0 updated to use native custom ids now.

there are some tests failing however that's expected, since they relate to deletion of files.

when a file is deleted in UT, it is not instantly deleted but it may take up to 30 seconds before it is deleted. some of the tests here assume that calling `hasItem(key)` directly after `deleteItem(key)` should return false, which is not the case for uploadthing.

idk how you wanna pursue given that knowledge. alter the tests? custom test suite for UT?

src/storage.ts
```diff
@@ -231,7 +231,7 @@ export function createStorage<T extends StorageValue>(
     async setItems(items, commonOptions) {
       await runBatch(items, commonOptions, async (batch) => {
         if (batch.driver.setItems) {
-          await asyncCall(
+          return asyncCall(
```
juliusmarminge, line 234: fairly certain this was a bug before. you called setItems and then setItem for every file so files were getting uploaded twice. 

noticed this since the newly introduced customids on uploadthing must be unique and we throw if that's not the case and we got duplicate requests when testing the setItems function before i changed the keyword here

pi0:
👀 canary

src/storage.ts
pi0: Nice catch. Do you mind splitting it into a new PR as well? 🙏🏼

pi0:
Exciting updates 🚀 Sorry i'm quite busy still will try to come to this PR as soon as could also for failing tests.

src/storage.ts
juliusmarminge: https://github.com/unjs/unstorage/pull/392

juliusmarminge (commented):
TODO: bump deps once properly released

pi0:
Hi dear @juliusmarminge and @markflorkowski 

Finally back on this. I've updated client to v7 impl and made more reactors with base key handling, etc. customId is a really interesting idea and things _almost_ work now!

Tests updated to run against a real (sandbox) endpoint. you need to update `.env` and manually run `pnpm vitest test/drivers/uploadthing.test.ts` (GH actions CI is ready and works after merge).


---

I found an interesting issue that seems happening in prod when listing keys. When a file with custom id is created and then deleted, any attempt to recreate it, it won't be returned from list anymore (!) ... which breaks `getKeys` test and `initial` test also if we remove random prefix in test file, might need you attention.

When that fixed, one last step is that we default to private ACL as this is the main purpose of unstorage for private server data.

juliusmarminge:
I tihnk what you're running into is this: https://github.com/pingdotgg/uploadthing/issues/948

docs/2.drivers/uploadthing.md
```diff
@@ -0,0 +1,32 @@
...
+Learn more about UploadThing.
+::
+
+```js
+import { createStorage } from "unstorage";
+import uploadthingDriver from "unstorage/drivers/uploadthing";
+
+const storage = createStorage({
+  driver: uploadthingDriver({
+    // apiKey: "<your api key>",
```
juliusmarminge, line 15:
```suggestion
    // token: "<your token>",
```

pi0:
Thinking of moving forward and iterating.

- Added note in docs that driver is experimental
- Added note in the docs about the known issue of same-key
- Used a "cloud-like" icon for docs until there is an icon in [iconify](https://icon-sets.iconify.design/) for uploadthing

juliusmarminge:
That sounds good to me. I'll try to remember and come back here when we've removed that limitation

PR #395, opened 2024-02-07
fix: `getItems`, `setItems` types

harlan-zw:
### ❓ Type of change

### 📚 Description

The type isn't using the generic so the types are broken when augmenting with a generic.

image

For consistency with `getItem` we also allow augmenting at a function level `getItems<{ foo: 'bar' }>`.

Also seems like the `value` was only accepting `string` for `setItems` whcih I've included a fix for also.

### 📝 Checklist


pi0 (approved):
Thanks!

PR #402, opened 2024-03-04
feat: add `keys`, `get`, `set`, `has` and `del` aliases

Atinux:
### ❓ Type of change

### 📚 Description

For a sexier usage :)

### 📝 Checklist

- [x] I have added a test


src/types.ts
```diff
@@ -125,4 +125,10 @@ export interface Storage<T extends StorageValue = StorageValue> {
     base?: string,
     options?: { parents?: boolean }
   ) => { base: string; driver: Driver }[];
+  // Aliases
+  keys: Storage["getKeys"];
+  get: Storage["getItem"];
+  set: Storage["setItem"];
+  has: Storage["hasItem"];
+  del: Storage["removeItem"];
```
pi0, line 133: How do you think about `remove` as alternative. This way we prevent abbreviation same as other new shortcuts

Atinux: Remove sounds good

Atinux: Added the `remove`, I like the `del` also as similar to Redis

pi0 (approved):

PR #407, opened 2024-03-08
refactor(netlify-blobs): update to v7

ascorbic:
### 🔗 Linked issue

#406 updates the dependency without fixing the imports in tests
### ❓ Type of change

### 📚 Description

Netlify Blobs are coming out of beta, and the latest version of the client has breaking changes to the local server used in tests. This PR updates the client and changes the imports for the tests. It also updates the docs to remove the reference to beta and add usage information for edge functions.

### 📝 Checklist


docs/2.drivers/netlify.md
```diff
@@ -25,7 +21,25 @@ const storage = createStorage({
 });
 ```
 
-You can create a deploy-scoped store by settings `deployScoped` option to `true`. This will mean that the deploy only has access to its own store. The store is managed alongside the deploy, with the same deploy previews, deletes, and rollbacks.
+You can use Unstorage in a Netlify edge function without installing any dependencies:
+
+```js
+import { createStorage } from "https://esm.sh/unstorage";
```
pi0, line 27: I would very love to advocate URL import patterns but it would be inconsistent to only do it here and also it probably deserves more explanation docs that for example Node.js/bundlers don’t support it out of the box at least.

If you are happy shall we defer adding this section to another PR?

docs/2.drivers/netlify.md
```diff
@@ -43,7 +57,7 @@ To use, you will need to install `@netlify/blobs` as dependency or devDependency
 ```json
 {
   "devDependencies": {
-    "@netlify/blobs": "*"
+    "@netlify/blobs": "latest"
```
pi0, line 60: Just checking, you would be okay opting users to use any latest major instead of documenting the tested latest major like ^1?

docs/2.drivers/netlify.md
ascorbic: This example is specific to Netlify Edge Functions though. They need this syntax if they're not being bundled by a framework because subpath imports from node modules don't work. I can see if I can find some wording that's clearer.

pi0: A more detailed explanation would be cool! BTW for Nuxt/Nitro users we do support it without CDN (only)

ascorbic: Yeah, it would work with anything that bundles it. It's just when doing a raw import, and is a limit of our current npm handling. It even works fine in this project when I manually import the compiled file. How's the updated wording?

docs/2.drivers/netlify.md
ascorbic: We want them to use >=7 because of changes in the way keys are namespaced. I'll add a section explaining this.

package.json
```diff
@@ -103,7 +103,7 @@
     "@azure/keyvault-secrets": "^4.8.0",
     "@azure/storage-blob": "^12.17.0",
     "@capacitor/preferences": "^5.0.7",
-    "@netlify/blobs": "^6.5.0",
+    "@netlify/blobs": "^6.5.0 || ^7.0.0",
```
ascorbic, line 106: I've expanded the peer dep so it's not a breaking change for users

anthonyholmes:
Looking forward to having this merged in! Do we think soon?

Issue #409, opened 2024-03-12
Optional ioredis


alexfriesen:
### Describe the feature

I would like to use this package but, it adds `ioredis` as a direct dependency. I suggest to set `ioredis` as a optional `peerDependency` instead.

### Additional information

- [X] Would you be willing to help implement this feature?

PR #410, opened 2024-03-13
fix(pkg): make `ioredis` dependency optional

alexfriesen:
### ❓ Type of change

- [x] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

### 📚 Description

Currently ioredis will be installed, even if you are not using it. With this change ioredis will be an optional peerDependency and you have to manually install it. 
Resolves: #409 

### 📝 Checklist


pi0 (approved):
Thanks and nice spot! While this is a breaking change but also makes sense for the sake of other driver usages. It was mainly my mistake that didn't spot this issue when merging initial PR.

Issue #435, opened 2024-05-01
Allow to return a blob for `getItemRaw` (http driver)


manniL:
### Describe the feature

To fetch binary files (e.g. images), using a blob response would be helpful for the http driver

A simple and optional implementation could be:

```diff
diff --git a/drivers/http.mjs b/drivers/http.mjs
index 99a72ce2289532f7bbc52ad6ff7ec5b8755aa62c..a07d5fe496341b4e1a8e68cdd0b4cd7154270c41 100644
@@ -32,7 +33,8 @@ export default defineDriver((opts) => {
           accept: "application/octet-stream",
           ...opts.headers,
           ...topts.headers
-        }
+        },
+        responseType: topts.responseType ?? opts.responseType
       }).catch(catchFetchError);
       return value;
     },
```

### Additional information

- [x] Would you be willing to help implement this feature?

Referencing #142 for this too :)

pi0:
Please issue issues to explain **why** not **how** 🙏🏼 (eg: a user code example or reproduction)

manniL:
@pi0 Of course!

**Why**: Because if you want to fetch binary files, such as *images* from the http server `unstorage` provides, there is no way to do so with the HTTP driver and ensure the formatting is correct.

The goal would be doing something like this in the browser (or whereever):

```ts
const storage = createStorage({
  driver: httpDriver({ base: "/api/storage" }),
});

const getItem = async () => {
  const blob = await storage.getItemRaw(name.value, {
    responseType: 'blob'
  })
  const url = URL.createObjectURL(blob)
  // do more with the URL
}
```

pi0:
Thanks for explaining. 

I think it makes sense but we should consider implementing a driver-agnostic `type` (ideally via [undio](https://undio.unjs.io) supported types) rather than patching for only one driver. (HTTP driver then can respect this type and prefer native `blob` or `response` return when it can preserve blob/file information) 

`responseType` is specific to the underlying library of HTTP driver (ofetch) and we cannot rely on it nor diverge the API only for one driver support (HTTP) because it can change and unstorage API should be consistent regardless of the driver.

> If you have a better idea of solving this problem,

In the meantime, I would suggest you directly use fetch for the HTTP driver if need blob access. `raw` support is still marked as experimental for unstorage after all for the very same reason of limitations and inconsistent behavior. 

Let's track this via https://github.com/unjs/unstorage/issues/142 👍🏼 (I have updated undio note in there)

manniL:
[...]

PR #436, opened 2024-05-01
feat(http): allow changing responseType

manniL:
### 🔗 Linked issue

#435 

### ❓ Type of change

### 📚 Description

This PR allows customizing the response type, ensuring that binary files can be sent as blob

Happy for initial feedback, will add tests + docs ☺️ 

### 📝 Checklist

### DIFF
--- a/src/drivers/http.ts
+++ b/src/drivers/http.ts
@@ -46,7 +46,7 @@ export default defineDriver((opts: HTTPOptions) => {
           ...opts.headers,
           ...topts.headers,
         },
-        responseType: topts.responseType ?? opts.responseType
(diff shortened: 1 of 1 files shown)


pi0:
Thanks but there is an open discussion and I'm not sure introducing a driver-specific option for return type would be best idea for future compatibility.

manniL:
@pi0 How else would you solve *fetching* blobs using the http driver (from the http server) then?

pi0:
Please use the issue first to showcase your usage, and then we can come up with a solution that is future-proof.

manniL:
@pi0 Of course! Following up in https://github.com/unjs/unstorage/issues/435

Issue #441, opened 2024-05-08
support raw methods for azure-storage-blob 


blowsie:
### Environment

Any / All

### Reproduction

https://github.com/unjs/unstorage/blob/main/src/drivers/azure-storage-blob.ts#L96

### Describe the bug

Calling `.toString()` [here](https://github.com/unjs/unstorage/blob/main/src/drivers/azure-storage-blob.ts#L96) is problematic, as the `getItem` function can longer be used with buffers. 

My suggestion would be to allow a type to be passed to the `getItem()` fn, or remove the `toString()` call and allow developers to call that manually, when they choose.

pi0:
Hi. We have `getItemRaw` that can be implemented and return the buffer/raw value. PR welcome to add that.

blowsie:
Will do thanks!

refs #https://github.com/unjs/unstorage/issues/142

PR #442, opened 2024-05-08
feat(azure-storage-blob): add raw support

blowsie:
### 🔗 Linked issue
#441 

### ❓ Type of change

### 📚 Description

Adds support for `getItemRaw` in `azure-storage-blob` 

### 📝 Checklist

Functions are not documented


pi0:
Thanks for PR!

- I think most of get/raw logic can now be shared using a function and we call toString for not raw version
- We need to also implement the `setItemRaw` method otherwise core will stringify when setting but not deserialize in getRaw
- Can you help to add an additional test to [driver test](https://github.com/unjs/unstorage/blob/main/test/drivers/azure-storage-blob.test.ts) that covers this? (/cc @itpropro he might be able to help 🙏🏼 )

blowsie:
> Thanks for PR!
> 
> * I think most of get/raw logic can now be shared using a function and we call toString for not raw version

I wasn't sure about extracting and reusing the fn, since when I reviewed all the other `getItemRaw` implementations there is no re-use. 
For this reason, I tried to copy the existing style.

> * We need to also implement the `setItemRaw` method otherwise core will stringify when setting but not deserialize in getRaw

I unfortunately don't know enough about how to do that, but can do some digging at some point if it helps.

> * Can you help to add an additional test to [driver test](https://github.com/unjs/unstorage/blob/main/test/drivers/azure-storage-blob.test.ts) that covers this? (/cc @itpropro he might be able to help 🙏🏼 )

Adding tests, would be a separate issue and not a blocker for this change, right?

pi0:
For setItemRaw, it _seems_ that our driver implementation will be the same. But we need to set it. see [here](https://github.com/unjs/unstorage/blob/main/src/storage.ts#L268) for what happens.

peterbud:
Hello team, is there any plans to complete this PR, or should I open a new one if I'd like to have this functionality?

pi0:
Merged via #565. Thanks for PR ❤️

Issue #396, opened 2024-02-07
`getItems` does not work with `prefixStorage`


harlan-zw:
### Environment

1.10.1

### Reproduction

https://stackblitz.com/edit/stackblitz-starters-rrstqp?file=index.js

### Describe the bug

Working

```ts
useStorage().getItems(['namespace:key'])
```

Not Working

```ts
prefixStorage(useStorage(), 'namespace').getItems(['key'])
useStorage('namespace').getItems(['key'])

// { key: "key", value: null }
```

It seems to switch back to the default driver (`memory`). I'm guessing this is because the mount is being discovered from the input key, without the base being applied first.

pi0:
Can you please make a sandbox ideally with unstorage only? 🙏🏼

harlan-zw:
Sorry I thought it was self-evident but it seems like something weird is going on.

Check this repro: https://stackblitz.com/edit/stackblitz-starters-rrstqp?file=index.js

The prefixStorage is not even working here :thinking: This works in Nitro though

harlan-zw:
Yes, sorry thought it was clear enough. Here is a repro: https://stackblitz.com/edit/stackblitz-starters-rrstqp?file=index.js

Would be great to add a starter to readme for quickly spinning one up.

dimasxp:
useStorage('namespace').clear() also not working with namespace

Mahdhir:
Same issue here.
prefix storage clear doesn't work
Also clear with base doesn't work for in-memory storage

PR #455, opened 2024-06-19
fix: make setItems and getItems work with prefixStorage

AQian0:
🔗 Linked issue

fix https://github.com/unjs/unstorage/issues/396

❓ Type of change

📚 Description

When using setItems and getItems, prefixStorage can't handle base and key splice properly. storageKeyProperties doesn't contain setItems and getItems, and the original processing doesn't support passing parameters as arrays.

I added a regular expression to determine what type the current storageKey should be, and when it meets the condition, it will use traversal to splice the base and key together. The regular expression is now set to /(set|get)\w+s/, which I think should be backward compatible with APIs that conform to the naming convention, and since this processing is performed during traversal, it can be overridden later if the developer isn't happy with the established processing.

📝 Checklist

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -5,8 +5,10 @@ type StorageKeys = Array<keyof Storage>;
 const storageKeyProperties: StorageKeys = [
   "hasItem",
   "getItem",
+  "getItems",
   "getItemRaw",
   "setItem",
+  "setItems",
   "setItemRaw",
   "removeItem",
   "getMeta",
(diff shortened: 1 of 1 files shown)


pi0:
Thanks for PR dear @AQian-Cup. However it is the right fix, it also changes behavior that can potentially break projects relying on current wrong behavior, therefore we need to land it as semver-major version.

pi0:
#611 should have this covered, thanks again for PR (sorry missed it from past) 

Considering [getItems/setItems](https://github.com/unjs/unstorage/blob/main/src/types.ts#L40-L51) were marked as experimental and it is a long lasting issue, releasing in semver-minor.

PR #459, opened 2024-07-05
fix(cf-kv): go through all pages to list the keys

farnabaz:
Based on the official documentation, KV list function has maximum limit of 1000 keys. I n order to get all keys from the KV binding, we need to go through all pages using the `cursor` property.

This PR updates `getKeys` function to leverage `cursor`.

Screenshot 2024-07-05 at 14 15 37

Originally raised in nuxt-hub/platform#254

### DIFF
--- a/src/drivers/cloudflare-kv-binding.ts
+++ b/src/drivers/cloudflare-kv-binding.ts
@@ -18,8 +18,18 @@ export default defineDriver((opts: KVOptions) => {
   async function getKeys(base: string = "") {
     base = r(base);
     const binding = getKVBinding(opts.binding);
-    const kvList = await binding.list(base ? { prefix: base } : undefined);
-    return kvList.keys.map((key) => key.name);
+    const keys = [];
+    let cursor: string | undefined = undefined;
+    do {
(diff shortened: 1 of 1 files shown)


src/drivers/cloudflare-kv-binding.ts
```diff
@@ -18,8 +18,18 @@ export default defineDriver((opts: KVOptions) => {
...
     base = r(base);
     const binding = getKVBinding(opts.binding);
-    const kvList = await binding.list(base ? { prefix: base } : undefined);
-    return kvList.keys.map((key) => key.name);
+    const keys = [];
+    let cursor: string | undefined = undefined;
+    do {
+      const kvList = await binding.list({ prefix: base || undefined, cursor });
+
+      keys.push(...kvList.keys.map((key) => key.name));
```
pi0, line 26: perf: i think we can do the map in last `return` line

farnabaz: Updated

pi0 (approved):
Thanks! (release might be in next please since there are other unreleased features. please ping me if was urgent to use hotfix)

Issue #462, opened 2024-07-08
Missing Favicon


moshetanzer:
### Environment

web

### Reproduction

unstorage.unjs.io

### Describe the bug

There is no favicon on this amazing site. Literally took me 3 mins to get in memory storage setup! Thanks so much. Just opening this for pr.

pi0:
Thanks for report again. Latest deployment has proper meta tag

PR #463, opened 2024-07-08
docs: add favicon

moshetanzer:
Resolves https://github.com/unjs/unstorage/issues/462


pi0:
Thanks for the PR. While it is missing, we should fix it in undocs to auto generate.

PR #470, opened 2024-07-31
feat(cloudflare-kv-binding): support `ttl` for `setItem`

barberj:
Nitro calls `setItem` with key and an entry object. The entry object has keys value, integrity, mtime, and expires. Set ttl on TransactionOptions from value before calling the driver's setItem.

resolves https://github.com/unjs/unstorage/issues/32 & [Nitro#2510](https://github.com/unjs/nitro/issues/2510)


src/storage.ts
```diff
@@ -223,6 +223,9 @@ export function createStorage<T extends StorageValue>(
       if (!driver.setItem) {
         return; // Readonly
       }
+      if (value && isPureObject(value) && value.expires) {
```
pi0, line 226: We should not assume any structure for `value` it is user-data.

src/drivers/cloudflare-kv-binding.ts
```diff
@@ -48,8 +49,12 @@ export default defineDriver((opts: KVOptions) => {
     },
     setItem(key, value, topts) {
       key = r(key);
+      let expirationOpts: TransactionOptions = {};
+      if (topts.ttl) {
+        expirationOpts.expirationTtl = topts.ttl;
+      }
       const binding = getKVBinding(opts.binding);
-      return binding.put(key, value, topts);
+      return binding.put(key, value, expirationOpts);
```
pi0, line 57: We were already passing all topts. this is a breaking change since limits what we set now. (as proper todo, we should _transform_ `ttl` option to `expirationTtl`.

pi0 (changes requested):

Issue #264, opened 2023-07-05
Vercel Blob Support


pi0:
### Describe the feature

https://vercel.com/docs/storage/vercel-blob

https://www.npmjs.com/package/@vercel/blob

Client: https://github.com/vercel/storage/blob/main/packages/blob/src/index.ts

We might either  use client or adopt it to use universal [node-fetch-native](https://github.com/unjs/node-fetch-native) + [uncrypto](https://github.com/unjs/uncrypto)

PS: I am still on the waitlist to try!

### Additional information

- [X] Would you be willing to help implement this feature?

PR #472, opened 2024-08-12
feat: add `vercel-blob` driver

PikiLee:
resolves https://github.com/unjs/unstorage/issues/264


ryanweston:
Is there anything I can do to help get this merged? @pi0

MickL:
@pi0 Can this be merged or is anything missing? Vercel is VERY popular and they have their own file storage [Vercel Blob](https://vercel.com/docs/storage/vercel-blob)

pi0:
Vercel blob uses upstash redis and [upstash driver](https://unstorage.unjs.io/drivers/upstash)  is available on [nightly channel](https://github.com/unjs/unstorage#nightly-release-channel) (via #500) and soon the next release, which gives same functionality.

While Vercel blob is still in Beta, there is a chance that it will be merged to [upstash](https://vercel.com/marketplace/upstash) directly.

<small>(PR, please read [this](https://github.com/pi0/tired-maintainer) before asking for "merge" 🙏🏼)</small>

pi0:
Thanks for PR. I've pushed a few refactors and improvements.

While tests pass against prod, what really concerns me is that we have to make every piece of data publicly accessible, and implementation also requires that we disable random suffixes.

I've added a warning to docs and runtime also but still users might mistakenly leak private info as only and only this driver on server-side has this behavior today. (netlify Blob for example in contrast, AFAIK encrypts and makes them always private).

I will contact vercel team as followup but in the meantime, anyone can directly use `@vercel/blob` in their projects and there is KV storage as well powered by upstash.

pi0:
Update: Private access will eventually be supported. 

For now updated driver to require `access: "public"` config explicitly to make sure devs know the fact.

pi0 (approved):

Issue #400, opened 2024-02-29
Generic SQL driver with db0


pi0:
Initial supported added via https://github.com/unjs/unstorage/pull/476 (thanks to @zsilbi ❤️)

Driver is currently available via [nightly channel](https://github.com/unjs/unstorage#nightly-release-channel) and any feedbacks are more than welcome!

Docs: https://unstorage.unjs.io/drivers/database

---

For supporting more SQL-based providers, we shall make a generic [db0](https://github.com/unjs/db0) based driver that works for all supported ones.

**What's the point of using SQL as KV?** I personally believe only to **reuse** same already setup service for another purpose that requires a simple KV interface. Otherwise, SQL is interface to interact with database and DB0 can be used standalone.

---

Today, only we have [planetscale](https://unstorage.unjs.io/drivers/planetscale) which is SQL based and I think we can move it too (and handle differences like "cache boost" mode here or could be there too.)

Celibioux:
What About [SurrealDB](https://github.com/surrealdb/surrealdb)?

pi0:
@Celibioux I have listed it for db0 tracker list: https://github.com/unjs/db0/issues/32

danielwaltz:
Love seeing this! One extra perk I'll throw out there to having this feature (that admittedly I got from watching a recent interview with DHH by Aaron Francis - relevant part [here](https://youtu.be/0rlATWBNvMw?t=2239)) is the ability to potentially use separate SQLite databases for things like a caching layer or queues.

I know DHH is kind of a controversial voice in the web dev space right now, and while I don't agree with all his viewpoints, I quite like the stance on using SQLite as a default for storage without needing to set up and connect to external services. I know this is the case with Nitro + db0 which I very much align with. Plus I find the argument that the cost for memory is significantly higher than the cost for storage without as large a performance benefit as it used to be thanks to SSDs very compelling.

After watching that part of interview, I immediately thought of unstorage and went looking to see if SQLite was supported! The rabbit hole landed me here. Just wanted to share this perspective and that I'm looking forward to seeing this land!

pi0:
Thanks for the feedback dear @danielwaltz. SQLite and other SQL dialects will for sure be supported in unstorage so you can leverage the same DB-File for the Nitro caching layer as well if you wish.

---

[...]

PR #476, opened 2024-09-04
feat: add experimental `db0` driver

zsilbi:
Hello!

Resolves https://github.com/unjs/unstorage/issues/400

I added tests for `libsql` (local file) and `postgresql` (only runs when the connection string in the `.env` is present.
And I also updated the docs.

Notes:
<details>
<summary>Inlining db0 in the vitest config was necessary because of this error from the postgres test</summary>
<pre>
FAIL  test/drivers/db0.test.ts [ test/drivers/db0.test.ts ]
SyntaxError: Named export 'Client' not found. The requested module 'pg' is a CommonJS module, which may not support all module.exports as named exports.
CommonJS modules can always be imported via the default export, for example using:

import pkg from 'pg';
const { Client } = pkg;
</pre>
</details>

Related: https://github.com/unjs/db0/issues/62 https://github.com/unjs/db0/pull/63


pi0:
Thanks for the PR dear @zsilbi it is a good start.

TBH i have to go over it for next steps (and we need a newer db0 release since it is stall in current experimental version) but some pointers if you can help in the meantime:

- We need to most importantly support `sqlite` dialect for db0 and have it as default
- Table creation should be automated too so that this driver can be used in zero config setups (we might need to have migration too later as a note)
- (chore) i would remove dotenv requirement. pglite and sqlite can be used for testing

src/drivers/db0.ts
```diff
@@ -0,0 +1,84 @@
...
+
+const DRIVER_NAME = "db0";
+
+export default defineDriver((opts: Db0DriverOptions) => {
+  opts.table = opts.table || "storage";
+
+  const dialects: Partial<Record<Dialect, Partial<Driver>>> = {
+    mysql: {
+      setItem: async (key, value) => {
+        await opts.database
```
pi0, line 30: Refactor: To improve maintenance, i would use an if condition inside main `setItem` to override behavior for `mysql` dialect

zsilbi:
>  Table creation should be automated too so that this driver can be used in zero config setups

Should it always try to create the table before the first query, or only after the first failed query and then retry?

To make it more easier to configure, `dialect` entries could also be added to `db0` connectors.
The database instance could return the dialect of the provided connector so this driver could use that.

pi0:
>  or only after the first failed query and then retry?

i wasn't thinking of it sounds a nice idea!

> To make it easier to configure, dialect entries could also be added to db0 connectors.

i was thinking of it actually! feel free to open an issue in db0 to support this.

zsilbi:
> > or only after the first failed query and then retry?
> 
> i wasn't thinking of it sounds a nice idea!

I got this working. It also stores the promise of the `CREATE TABLE` query, so concurrent queries could queue up waiting for the same promise to be fulfilled.

> I would remove dotenv requirement. pglite and sqlite can be used for testing

I'll change it to use pglite when it will be available: https://github.com/unjs/db0/pull/110

PR #483, opened 2024-09-11
fix: allow omitting transaction options

danielroe:
issue introduced in https://github.com/unjs/unstorage/pull/250 requires an `opts` argument to be passed to all drivers

this may be intentional so feel free to close if so!

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -33,7 +33,10 @@ export interface Driver<OptionsT = any, InstanceT = any> {
     commonOptions?: TransactionOptions
   ) => MaybePromise<{ key: string; value: StorageValue }[]>;
   /** @experimental */
-  getItemRaw?: (key: string, opts?: TransactionOptions) => MaybePromise<unknown>;
+  getItemRaw?: (
+    key: string,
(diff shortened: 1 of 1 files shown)


src/types.ts
```diff
@@ -33,11 +33,14 @@ export interface Driver<OptionsT = any, InstanceT = any> {
     commonOptions?: TransactionOptions
   ) => MaybePromise<{ key: string; value: StorageValue }[]>;
   /** @experimental */
-  getItemRaw?: (key: string, opts: TransactionOptions) => MaybePromise<unknown>;
+  getItemRaw?: (
+    key: string,
+    opts?: TransactionOptions
```
pi0, line 38: [core sets it](https://github.com/unjs/unstorage/blob/main/src/storage.ts#L164) when calling drivers.

If in `Storage` it is not optional i think that is the mistake

pi0:
@danielroe Did you remember where occured with this issue? I just checked only types are marking it as "set" in `Driver` hooks (as core sets it) and optional in `Storage` interface. 

Can we close?

PR #486, opened 2024-09-15
refactor(netlify): support @netlify/blobs v8 and improve options types

### LINKED ISSUE #485: @netlify/blobs@8 is compatible but is reported as a peer dependency error
### Environment

```
$ node -v
v20.8.0
$ pnpm -v
9.6.0
```

### Reproduction

1. create a site, e.g. `pnpm create solid`
2. add unstorage: `pnpm add @unjs/unstorage@latest`
3. add @netlify/blobs: `pnpm add @netlify/blobs@latest`
    - this reports a peer dependency error (some other package managers might bail here)
5. use the Netlify unstorage driver
6. it works :)

### Describe the bug

That's about it. PR coming in a minute.

### Additional context

_No response_

### Logs

```sh
 WARN  Issues with peer dependencies found
.
└─┬ vinxi 0.4.3
  └─┬ unstorage 1.10.2
    └── ✕ unmet peer @netlify/blobs@"^6.5.0 || ^7.0.0": found 8.0.1
```
```

### PR DESCRIPTION
**Pull request by serhalp**
The peer dependency only specifies support for v6 and v7, but v8 works just fine.

Closes #485.

This adds v8 to the peer dependencies, bumps the dev dep to v8, updates the tests (there's a new requirement that's specific to unusual use cases like unstorage's tests, which are run outside of the Netlify platform and the Netlify CLI), and adjust the driver types to ensure compatibility with all supported versions.

See https://github.com/netlify/blobs/pull/183.

I believe @netlify/blobs@8 will error in local dev when using netlify-cli before 17.21.1. This doesn't seem like unstorage's problem though, so I didn't do anything about that here.

### CODE REVIEW

File: src/drivers/netlify-blobs.ts [hunk 1]
```diff
@@ -33,13 +33,19 @@ export interface NetlifyDeployStoreOptions extends NetlifyBaseStoreOptions {
   deployID?: string;
 }
 
+export interface NetlifyDeployStoreV8Options extends NetlifyDeployStoreOptions {
+  // TODO(serhalp) Export this type from @netlify/blobs instead
```
serhalp, line 37: I'll leave this PR as a draft until I address this.

test/drivers/netlify-blobs.test.ts
```diff
@@ -39,6 +39,8 @@ describe("drivers: netlify-blobs", async () => {
       token,
       siteID,
       deployID: "test",
+      // Usually defaulted via the environment; only required in a test environment like this
+      region: "us-east-1",
```
pi0, line 43: Is there a fallback behavior if there is no env & default set?

serhalp: No, it bails. But that should only be possible when running outside of the Netlify platform and outside of the Netlify CLI (where Netlify Blobs wouldn't work anyway... so you wouldn't even get to here), or like I mentioned in the PR description if you're using an old version of the Netlify CLI (<17.21.1).

I'll look into making that error more actionable though. It should ideally tell users to upgrade to netlify-cli@latest if they're running through the CLI, instead of just saying `region` is missing.

src/drivers/netlify-blobs.ts
serhalp: Done!

test/drivers/netlify-blobs.test.ts
serhalp: Done!

src/drivers/netlify-blobs.ts
```diff
@@ -5,40 +5,42 @@ import type {
...
   /** If set to `true`, the store is scoped to the deploy. This means that it is only available from that deploy, and will be deleted or rolled-back alongside it. */
   deployScoped?: boolean;
-  /** Required during builds, where it is available as `constants.SITE_ID`. At runtime this is set automatically. */
-  siteID?: string;
-  /** Required during builds, where it is available as `constants.NETLIFY_API_TOKEN`. At runtime this is set automatically. */
-  token?: string;
-  /** Used for advanced use cases and unit tests */
-  apiURL?: string;
-  /** Used for advanced use cases and unit tests */
-  edgeURL?: string;
```
pi0, line 27: Removal of this, i think breaks types for v7 users (in peer deps, we still allow this)

pi0 (approved):
Thank you ❤️

PR #487, opened 2024-09-17
docs: missing values return null, not undefined

ralacerda:
Documentation is saying that missing values are `undefined`, but they are in fact `null`.


pi0 (approved):

pi0:
Thanks.

PR #495, opened 2024-10-10
feat: add `null` driver

### LINKED ISSUE #217: `null` driver
Similar to unix [/dev/null](https://en.wikipedia.org/wiki/Null_device) to make mount points no-op

ToBinio:
resolves #217 

I might have misunderstood what was meant by "null driver".
But a driver that simply does not store data would be great for temporarily deactivating caching within Nitro.


docs/2.drivers/null.md
```diff
@@ -0,0 +1,20 @@
+---
+icon: bi:trash3-fill
+---
+
+# Null
+
+> Discards all data.
+
+::warning
+This driver does NOT store any data. It will discard any data written to it and will always return null!
```
pi0, line 10:
```suggestion
This driver does NOT store any data. It will discard any data written to it and will always return null similar to [`/dev/null`](https://en.wikipedia.org/wiki/Null_device)
```

pi0 (approved):

pi0:
Thanks!

PR #500, opened 2024-10-18
feat: add `upstash` driver

fahreddinozcan:
This PR adds Upstash Driver to drivers. Under the hood, it uses HTTP based @upstash/redis library, which differentiates it from the regular TCP based Redis driver. Also, vercel-kv is soon to be deprecated, and it's currently uses @upstash/redis under the hood. The Upstash Driver also aims to replace it.


ascorbic:
Thanks @fahreddinozcan! This does need tests and docs adding before it can be merged. I think they can be based on the Vercel KV and/or Redis ones. It also needs adding to `src/index.ts`

src/drivers/upstash.ts
```diff
@@ -25,8 +25,9 @@ export default defineDriver<UpstashOptions, Redis>(
       if (redisClient) {
         return redisClient;
       }
-      redisClient = new Redis(options);
-
+      const url = options.url || process.env.UPSTASH_REDIS_REST_URL;
+      const token = options.token || process.env.UPSTASH_REDIS_REST_TOKEN;
```
ascorbic, line 29: This would mean it won't work in environments without `process.env`

ascorbic: Ah, looks like upstash/redis has a cross-platform static method `fromEnv`. Perhaps we could check if the options are defined, and if not then try to use that.

pi0 (approved):
Thanks!

PR #502, opened 2024-11-01
fix: revert chokidar to v3

pi0:
- https://github.com/nuxt/nuxt/issues/29744
- https://github.com/vitejs/vite/issues/18527

Revert Chokidar to v3 temporarily (updated in v1.13.0 via #489) due to downstream issues.

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,7 +45,7 @@
   },
   "dependencies": {
     "anymatch": "^3.1.3",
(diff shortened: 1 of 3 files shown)


talentlessguy:
won't setting `usePolling` to true help?

pi0:
It was a hotfix to avoid ecosystem-wise issues. polling has overhad.

Issue #277, opened 2023-07-24
String types are not preserved


olivbau:
### Environment

nodejs
v1.8.0

### Reproduction

```js
const storage = createStorage();
await storage.setItem('123', '456')
const item = await storage.getItem('123');
console.log(item, typeof item)
// 456, number
// instead of: "456", string
```

### Describe the bug

Storage convert string into number for no reason

EDIT:After investigation, it's probably due to destr
https://github.com/unjs/unstorage/blob/main/src/storage.ts#L173C9-L173C21

molvqingtai:
https://github.com/unjs/unstorage/blob/95da940f26f2b58dedb60e807d977f1a0a944a61/src/_utils.ts#L35
The error is caused by the String(value) method in stringify. Should the behavior of stringify be consistent with JSON.stringify?



String():
```js
String('1') // '1'
String(true) // 'true'
String(null) // 'null'
String('[]') // '[]'
```

JSON.stringify & JSON.parse :
```js
JSON.stringify('1') // '"1"'
JSON.stringify(true) // 'true'
JSON.stringify(null) // 'null'
JSON.stringify('[]') // '"[]"'


JSON.parse(JSON.stringify('1')) //  '1'
JSON.parse(JSON.stringify(true)) // true
JSON.parse(JSON.stringify(null)) // null
JSON.parse(JSON.stringify('[]')) // '[]'
```

idb-keyval:
```js
import { set, get } from 'idb-keyval'

await set('foo', '[]')
await get('foo') // '[]'

await set('bar', null)
await get('bar') // null

```
Inconsistent behavior of String() and JSON.stringify() will lead to confusion, which makes it impossible to migrate directly from idb-keyval to unstorage



I use [jsonr](https://github.com/perfsee/perfsee/tree/main/packages/jsonr) to store the compressed json into storage. Because the value obtained by getItem is not the value originally stored, an error is reported.

molvqingtai:
I tried using JSON.stringify to handle the string value, but this broke too many test cases

pi0:
Following up on #591, I think we can at least expose a new option for `setItem` as `setItem(key, value, { preserveTypes: true })` (or global unstorage option) and default it to `true` in v2.

PR #505, opened 2024-11-05
fix: The get value and set value are inconsistent. (#277)

molvqingtai:
Before the fix:
```js
await setItem('foo', '[]')
await getItem('foo') // []
```

After the fix:
```js
await setItem('foo', '[]')
await getItem('foo') // '[]'
```

This will fix: https://github.com/unjs/unstorage/issues/277

### DIFF
--- a/src/_utils.ts
+++ b/src/_utils.ts
@@ -31,11 +31,7 @@ function isPureObject(value: any) {
 }
 
 export function stringify(value: any): string {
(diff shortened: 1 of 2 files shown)


pi0:
Thanks for PR. As noticed it makes lots of tests fail since it is a breaking behavior change.

I will consider it for unstorage v2 👍🏼

Issue #508, opened 2024-11-14
Enhance ts type


wzc520pyfm:
### Describe the feature

We can improve the ts type as follows:
```ts
type TestObjType = {
  a: number;
  b: boolean;
};
type MyStorage = {
  foo: string;
  baz: TestObjType;
};
const storage = createStorage<MyStorage>();
await storage.getItem("foo"); // => <string>
await storage.getItem("baz"); // => <TestObjType>
storage.setItem("foo", "val"); // Check ok
storage.setItem("foo", 123); // TS error
```

and get hints as you type:
<img width="691" alt="image" src="https://github.com/user-attachments/assets/ad127e6c-c5fb-47f9-a9a3-ea45281320de">

### Additional information

- [X] Would you be willing to help implement this feature?

PR #509, opened 2024-11-14
feat: typed storage interface

wzc520pyfm:
Refer: #508 

We can improve the ts type as follows:
```ts
type TestObjType = {
  a: number;
  b: boolean;
};
type MyStorage = {
  foo: string;
  baz: TestObjType;
};
const storage = createStorage<MyStorage>();
await storage.getItem("foo"); // => <string>
await storage.getItem("baz"); // => <TestObjType>
storage.setItem("foo", "val"); // Check ok
storage.setItem("foo", 123); // TS error
```

and get hints as you type:
image


docs/1.guide/1.index.md
```diff
@@ -310,6 +310,20 @@ await storage.getItem("k"); // => <string>
 
 storage.setItem("k", "val"); // Check ok
 storage.setItem("k", 123); // TS error
+
+type TestObjType = {
+  a: number;
+  b: boolean;
+};
+type MyStorage = {
```
pi0, line 318: For future expansion of first generic, i prefer that we namespace it like `Storage<{ data: {} }>`

(also can you please add ts example to a new section in the end of page? 🙏🏼

wzc520pyfm: sorry, I don't understand, could you explain it more detail? I guess you mean to adjust the definition of getItem like:
```ts
type NsObject = {
  data: object
  [key: string]: any
}
export interface Storage<T extends StorageValue = StorageValue> {
  // ...
  getItem<U extends Extract<T, NsObject>, K extends keyof U["data"]>(
     key: K, ops?: TransactionOptions
  ): Promise<U["data"][K] | null>;
  getItem<U extends T>(key: string, opts?: TransactionOptions ): Promise<U | null>;
}
```

image

wzc520pyfm: I agree to use data as the namespace, I have updated.

docs/1.guide/1.index.md
```diff
@@ -348,3 +348,21 @@ In [strict mode](https://www.typescriptlang.org/tsconfig#strict), it will also r
...
 ```
+
+**Specifying namespace:**
+
+```ts
+type NsStorageValue = {
+  data: {
+    foo: string;
+    baz: number;
+  }
```
danielroe, line 359: what about `items`?

src/types.ts
```diff
@@ -61,13 +61,30 @@ export interface Driver<OptionsT = any, InstanceT = any> {
   watch?: (callback: WatchCallback) => MaybePromise<Unwatch>;
 }
 
+type NsStorageValue = {
+  data: unknown;
+  [key: string]: unknown;
+};
+
+type NsDataType<T extends NsStorageValue> = T["data"];
```
danielroe, line 69: I'm not super-excited about the `Ns` prefix. Really `StorageMeta` would have been perfect but it's already used. So maybe something like `StorageDefinition` and `StorageItemMap`?

danielroe (commented):
This is really nice to see!

I wonder if we could support extracting keys with `:` - so having a `namespace:child` would get the right types with a `base` of `namespace` and a key of `child`....

src/types.ts
pi0: What about TypedStorage?

danielroe: That could work! But in a sense `Typed` is true by definition as it's a type. So _what_ is being typed? I guess it partly depends on what else is in there, but `Definition` suggests that it totally describes the storage setup (which is why you wanted `data` as a key within it, right?)....

danielroe (approved):

pi0 (approved):
❤️

PR #512, opened 2024-11-17
refactor(indexedb): keep native json types in storage values

### LINKED ISSUE #511: Support raw storage for IndexedDB
### Describe the feature

The driver always stores objects in a stringified form, which makes it more difficult to inspect properties when storing large objects.

I made a workaround by creating a custom indexedDB driver replacing this line:
https://github.com/unjs/unstorage/blob/e42c01d0c22092f394f57e3ec114371fc8dcf6dd/src/drivers/indexedb.ts#L40-L42

with:
```ts
setItem(key, value) {
  try {
    value = JSON.parse(value)
  }
  catch { /* use string value */ }
  return set(makeKey(key), value, customStore);
},
```

Viewed in `Application` tab:
- Current behavior:
image

- Workaround behavior:
image

This workaround ensures that object values are stored as objects, otherwise they continue to be stored as strings.

### Additional information

ahmedrangel:
Resolves #511


pi0:
Thanks for PR. I think instead of changing driver behavior (and try parse overhead) we can simply implement `getItemRaw` and `setItemRaw` for indexed DB that passes `value` as is.

PR #517, opened 2024-12-07
cloudflareR2bindings: Fixed getItemRaw()

MickL:
TLDR: Files saved to Cloudflare R2 were not returned correctly by `.getItemRaw()`, this PR fixes that.

---

I tried to save files into Cloudflare R2 from a Cloudflare worker by a binding. Saving was no problem by use of `.setItemRaw()`, the file was added correctly to R2. But getting the item I had weird results:

* `.getItem()` returns a string, e.g. when the file was a pdf, the response was the markup of the pdf as a string. (btw this is because it is calling `.text()`, not sure if this is usefull at all)
* `.getItemRaw()` returns an empty object

I checked [the docs](https://developers.cloudflare.com/r2/api/workers/workers-api-usage/#4-access-your-r2-bucket-from-your-worker) and they do it like this:

```ts
export default defineEventHandler(async (event) => {
    const object = await globalThis.__env__.BINDING.get('key')`
    return object.data;
});
```

In this example above, `object` is of type [R2ObjectBody](https://github.com/cloudflare/workers-types/blob/master/index.d.ts#L992), [reference is here](https://developers.cloudflare.com/r2/api/workers/workers-api-reference/#r2objectbody-definition)

In the [current implementation](https://github.com/unjs/unstorage/blob/main/src/drivers/cloudflare-r2-binding.ts#L53) `object.arrayBuffer()` is returned instead of `.body`. I am not exactly sure why the author chose to do that, because:

1. It doesnt work (i tried to return image and pdf)
2. I would expect a "raw" function not to call a function that obviously does transformations

Therefor I changed `getItemRaw()` to return `.body` which now returns files correctly.

### DIFF
--- a/src/drivers/cloudflare-r2-binding.ts
+++ b/src/drivers/cloudflare-r2-binding.ts
@@ -50,7 +50,7 @@ export default defineDriver((opts: CloudflareR2Options = {}) => {
     getItemRaw(key, topts) {
       key = r(key);
       const binding = getR2Binding(opts.binding);
-      return binding.get(key, topts).then((r) => r?.arrayBuffer());
+      return binding.get(key, topts).then((r) => r?.body);
     },
     async setItem(key, value, topts) {
       key = r(key);


pi0:
I have pushed some better tests (#518) to validate `getItemRaw` behavior against actual platform behavior, and it works as expected. 

It is strange that you are getting an empty object. How do you test it? console.log might be misleading (please consider always sharing a reproduction this way we can help faster and better 🙏🏼)

`raw` methods return a raw **buffer** (ie: resolve value) and `.body` is a stream which is why. 

We could return a `Uint8Array` to be more consistent with other drivers but it will be a behavior change.

From this point:
- As an enhancement, we could support `.getItemRaw(key, { type: 'stream' }}` if you like to have stream type like this PR changes
- As a (possible) fix, we need a minimal reproduction that shows currently there is a bug. Addtional tests work as expected so it might not be a bug.

pi0:
~> To allow specify raw type: https://github.com/unjs/unstorage/pull/519

MickL:
What did you test exactly that arrayBuffer was correct? Did you upload a file to R2 using setItem?

I saved an uploaded file like this, and I only tested deployed to Cloudflare worker.

**Upload file:**

```ts
export default defineEventHandler(async (event) => {
  const files = await readMultipartFormData(event);

  await useStorage('files').setItemRaw('some-key', files[0].data);

  return 'ok';
});
```

**Get file:**

1. `getItem()` returns the text of a pdf
```
export default defineEventHandler(async (event) => {
  return useStorage('files').getItem('some-key');
});
```

Response:
```
�PNG


IHDR�*�W�biTXtXML:com.adobe.xmp
<?xpacket begin="﻿" id="W5M0MpCehiHzreSzNTczkc9d"?>
<x:xmpmeta xmlns:x="adobe:ns:meta/" x:xmptk="XMP Core 6.0.0">
```

3. `getItemRaw()` returns an empty object:

```
export default defineEventHandler(async (event) => {
  return useStorage('files').getItemRaw('some-key');
});
```

Response:
```json
{}
```

Btw. on localhost using memory driver this works with both: `getItem()` and `getItemRaw()`.

pi0:
Tests: https://github.com/unjs/unstorage/blob/main/test/drivers/cloudflare-r2-binding.test.ts

`getItemRaw('some-key')` returns an `ArrayBuffer` and I guess h3 does not support it as direct response (it handles Buffer and Buffer-like like `Uinit8Array`) you might want to do something like:

```js
const val = await useStorage('files').getItemRaw('some-key')
return new Uint8Array(val)
```

(another option is `return new Response(val)`)

Regardless there is no issue with unstorage driver itself and next release allows you to specify what raw type you want.

MickL:
Opened here: https://github.com/unjs/h3/issues/931

PR #519, opened 2024-12-10
feat(cloudflare-r2-binding): allow specify raw type

pi0:
(context: #517 by @MickL)

This PR allows `getItemRaw(key, { type: "..." })` with specified return type based on user needs: object/stream/blob/bytes/arrayBuffer are possible (check docs)

### DIFF
--- a/docs/2.drivers/cloudflare.md
+++ b/docs/2.drivers/cloudflare.md
@@ -163,6 +163,6 @@ const storage = createStorage({
 - `getItemRaw(key, { type: "..." })`
   - `type: "object"`: Return the [R2 object body](https://developers.cloudflare.com/r2/api/workers/workers-api-reference/#r2objectbody-definition).
   - `type: "stream"`: Return body stream.
-  - `type: "blob"`: Return data blobl.
+  - `type: "blob"`: Return a `Blob`.
   - `type: "bytes"`: Return an `Uint8Array`.
   - `type: "arrayBuffer"`: Return an `ArrayBuffer` (default)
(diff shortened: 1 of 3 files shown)


docs/2.drivers/cloudflare.md
```diff
@@ -157,3 +157,12 @@ const storage = createStorage({
 
 - `binding`: Bucket binding or name. Default is `BUCKET`.
 - `base`: Prefix all keys with base.
+
+**Transaction options:**
+
+- `getItemRaw(key, { type: "..." })`
+  - `type: "object"`: Return the [R2 object body](https://developers.cloudflare.com/r2/api/workers/workers-api-reference/#r2objectbody-definition).
+  - `type: "stream"`: Return body stream.
+  - `type: "blob"`: Return data blobl.
```
pi0, line 166:
```suggestion
  - `type: "blob"`: Return a `Blob`.
```

MickL:
Maybe

* getItemRaw() should return .body
* getItem() should have the option to set return type

?

What I just saw from the docs is that they return .body and I guess this is exactly what is needed to return the correct object? But I am not very familiar with the differences of ArrayBuffer, Blob and ReadableStream.

Example from Cloudflare: https://developers.cloudflare.com/r2/examples/demo-worker/

pi0:
Unstorage `raw*` methods return a **resolved value** and `.body` is a stream which is not resolved. All usages are valid for CF R2 bindings and it depends on how we want it to be consumed. for unstorage resolved makes sense (but this PR allows you to use other possible ways).

MickL:
I saw you marked type `bytes` to default for v2. Would then the following code work with H3? If yes, I am okay with the new implementation:

```ts
export default defineEventHandler(async (event) => {
  return useStorage('files').getItem('some-key');
});
```

pi0:
Yes it should. If you had time can you please make an issue + reproduction in h3 for supporting arrayBuffer too? (it looks an issue there)

PR #525, opened 2024-12-18
docs(netlify): add consistency option

ascorbic:
Updates the Netlify driver docs to add the `consistency` option, and the advanced option `uncachedEdgeURL`.


pi0 (approved):
Thanks!

Issue #531, opened 2024-12-19
`getItem` now returns `Promise<null>`


danielroe:
### Environment

latest version on renovate update

### Reproduction

https://github.com/nuxt/nuxt/pull/30298

### Describe the bug

Seems like a regression from https://github.com/unjs/unstorage/pull/509 - we need to fall back to `StorageValue`

PR #532, opened 2024-12-19
fix(types): fallback to `StorageValue` for un-typed stores

danielroe:
resolves https://github.com/unjs/unstorage/issues/531

we didn't catch the issue because `toMatchTypeOf` doesn't mean what it sounds like


pi0:
thnx. i think one type issue is in storage.ts

pi0 (approved):

PR #538, opened 2024-12-25
build: remove extra cli deps

bluwy:
`src/cli.ts` is no longer exported as a binary since https://github.com/unjs/unstorage/commit/5dd3f9bdb1dac4a7998acd5a3e15d17ab600f446. It uses `citty` and `listhen` which can both be moved as dev deps now.


pi0:
Oh nice find! Unstorage CLI drop was kinda unintended but considering it happened from 1.0.0 and no user demand I guess we can drop it.

pi0 (approved):

Issue #536, opened 2024-12-25
Generic of useStorage.getItem not work in 1.14.1


kingyue737:
### Environment

------------------------------
- Operating System: Linux
- Node Version:     v18.20.3
- Nuxt Version:     3.14.1592
- CLI Version:      3.17.2
- Nitro Version:    2.10.4
- Package Manager:  npm@10.2.3
- Builder:          -
- User Config:      default
- Runtime Modules:  -
- Build Modules:    -
------------------------------

### Reproduction

https://stackblitz.com/edit/github-5inkrb1i?file=server%2Ffoo.get.ts

### Describe the bug

v1.13.1 does not have this issue

open the reproduction and you will see type error:
```
No overload expects 1 type arguments, but overloads do exist that expect either 0 or 2 type arguments.
```
![image](https://github.com/user-attachments/assets/962149d6-a62f-473e-a245-ce786a204b95)

### Additional context

Upgrade nuxt to v3.15 will also upgrade unstorage to v1.14.1
For those who encounter this issue, you can override the version of unstorage in package.json:
```json
{
  "pnpm": {
    "overrides": {
      "unstorage": "1.13.1"
    }
  },
}
```

PR #543, opened 2024-12-26
fix(types): allow passing generic to untyped storage

danielroe:
resolves https://github.com/unjs/unstorage/issues/536


pi0 (approved):
Thanks!

PR #546, opened 2024-12-28
feat: add `depth` option to `getKeys`

### LINKED ISSUE #539: Support `depth` for `getKeys()`
### Describe the feature

hello there, im using unStorage to store images and files in filesystem im making it dynamic so can store files in different folders, but when i tried to get keys inside folder that have sub folders i got all items in sub folders too, i guess its expected behavior but would be great to have option to disable this behavior and get only current folder items

e.g.

```js
// incudeSubFolders: true
[
    "current.webp",
    "current.webp",
    "test:4.webp",
    "test:5.webp",
    "hello:world:1.webp",
    "hello:world:2.webp",
    "hello:world:3.webp"
]
```

```js
// incudeSubFolders: false
[
    "current.webp",
    "current.webp"
]
```

does that make sense? cause for now i need to filter keys and exclude all that have `:`
Thanks!

### Additional information

43081j:
Adds a new `depth` option to the `getKeys` signature.

Also implements that option in the following drivers:

- `fs`
- `fs-lite`
- `azure-storage-blob`

Fixes #539

Let me know if this is the kind of implementation you had in mind @pi0 

While I was wandering around the fs code, I did also get somewhat confused by `base` in `getKeys(base)`. It is probably me misunderstanding, but if I do `getKeys("subdir")`, I assumed I'd get all the files in `{root}/subdir` and below. However, it seems the current impl ignores `base` and hard codes `.` for traversal

Changing that to `base ?? "."` breaks some tests, though. it also results in some keys being dropped in the test driver/harness because it expects the keys to be `{base}{key}`, which isn't true in the fs driver it seems

### DIFF
--- a/src/drivers/azure-storage-blob.ts
+++ b/src/drivers/azure-storage-blob.ts
@@ -36,6 +36,47 @@ export interface AzureStorageBlobOptions {
 
 const DRIVER_NAME = "azure-storage-blob";
 
+async function getKeysByDepth(
+  client: ContainerClient,
+  depth: number
+): Promise<string[]> {
+  const queue: Array<{ depth: number; name: string }> = [];
+  let current: { depth: number; name: string } | undefined = {
+    name: "",
+    depth: 0,
+  };
+  const keys: string[] = [];
+
+  do {
+    const iterator = client
+      .listBlobsByHierarchy("/", {
+        prefix: current.name,
+      })
+      .byPage({ maxPageSize: 1000 });
+
+    for await (const result of iterator) {
+      const { blobPrefixes, blobItems } = result.segment;
+
+      if (blobPrefixes && current.depth < depth) {
+        for (const childPrefix of blobPrefixes) {
+          queue.push({
+            name: `${current.name}${childPrefix}`,
+            depth: current.depth + 1,
+          });
+        }
+      }
+
+      for (const item of blobItems) {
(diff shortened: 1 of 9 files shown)


pi0:
I appreciate the PR. For the first step, we need to implement core/generic functionality

- `storage.get(key, { depth }`
- In core (`storage.ts`) apply the filter (so even if the driver does not natively implement depth support behavior is consistent)

Before following up with adding driver support, I am thinking of exposing a new `capabilities: {}` object from drivers so the core is aware of the native ability of drivers. (we were inferring this for `raw` methods but this one needs meta)

Also, PR welcome to improve fs/-lite drivers please consider limiting changes in PRs to the scope to make review easier 🙏🏼

43081j:
so we need `get` to support `depth` just the same as `getKeys`?

> In core (storage.ts) apply the filter (so even if the driver does not natively implement depth support behavior is consistent)

how do you suggest we do this, given not all keys will be hierarchical

do we just assume any key with `/` in it is hierarchical? i.e. what delimiter do we default to?

pi0:
my bad i mean `getKeys`/`keys`(alias)

delimiter is normalized to `:` always.

43081j:
I have opened #547 so will close this one, as the followups are more likely to be one-pr-per-driver (once we know how `capabilities` will work)

PR #547, opened 2024-12-28
feat: support `maxDepth` for `getKeys`

43081j:
#539

This implements the `depth` option in `getKeys` at the storage layer (above drivers).

Roughly works by treating the `:` delimiter as a depth change and filters out those entries with a depth higher than has been requested.

In future, drivers can implement it themselves and we can make this logic conditional based on something like `capabilities.depth`

### DIFF
--- a/src/drivers/netlify-blobs.ts
+++ b/src/drivers/netlify-blobs.ts
@@ -1,10 +1,10 @@
 import { createError, createRequiredError, defineDriver } from "./utils";
+import { type GetKeysOptions } from "../types";
 import { getStore, getDeployStore } from "@netlify/blobs";
 import type {
   Store,
   BlobResponseType,
   SetOptions,
-  ListOptions,
   GetStoreOptions,
   GetDeployStoreOptions,
 } from "@netlify/blobs";
@@ -96,10 +96,7 @@ export default defineDriver((options: NetlifyStoreOptions) => {
     removeItem(key) {
       return getClient().delete(key);
     },
-    async getKeys(
-      base?: string,
-      tops?: Omit<ListOptions, "prefix" | "paginate">
-    ) {
+    async getKeys(base?: string, tops?: GetKeysOptions) {
       return (await getClient().list({ ...tops, prefix: base })).blobs.map(
         (item) => item.key
       );
--- a/src/drivers/netlify-blobs.ts
+++ b/src/drivers/netlify-blobs.ts
@@ -5,6 +5,7 @@ import type {
   Store,
   BlobResponseType,
   SetOptions,
+  ListOptions,
   GetStoreOptions,
   GetDeployStoreOptions,
 } from "@netlify/blobs";
@@ -96,7 +97,10 @@ export default defineDriver((options: NetlifyStoreOptions) => {
     removeItem(key) {
       return getClient().delete(key);
     },
-    async getKeys(base?: string, tops?: GetKeysOptions) {
+    async getKeys(
+      base?: string,
+      tops?: GetKeysOptions & Omit<ListOptions, "prefix" | "paginate">
+    ) {
       return (await getClient().list({ ...tops, prefix: base })).blobs.map(
         (item) => item.key
       );
--- a/src/storage.ts
+++ b/src/storage.ts
@@ -10,7 +10,12 @@ import type {
 } from "./types";
 import memory from "./drivers/memory";
 import { asyncCall, deserializeRaw, serializeRaw, stringify } from "./_utils";
-import { normalizeKey, normalizeBaseKey, joinKeys } from "./utils";
+import {
+  normalizeKey,
+  normalizeBaseKey,
+  joinKeys,
+  filterKeyByDepth,
+} from "./utils";
 
 interface StorageCTX {
   mounts: Record<string, Driver>;
@@ -362,11 +367,13 @@ export function createStorage<T extends StorageValue>(
           ...maskedMounts.filter((p) => !p.startsWith(mount.mountpoint)),
         ];
       }
-      return base
-        ? allKeys.filter(
-            (key) => key.startsWith(base!) && key[key.length - 1] !== "$"
-          )
-        : allKeys.filter((key) => key[key.length - 1] !== "$");
+      return allKeys
+        .filter((key) => filterKeyByDepth(key, opts.depth))
+        .filter(
+          base
+            ? (key) => key.startsWith(base!) && key[key.length - 1] !== "$"
+            : (key) => key[key.length - 1] !== "$"
+        );
     },
     // Utils
     async clear(base, opts = {}) {
--- a/src/storage.ts
+++ b/src/storage.ts
@@ -368,7 +368,7 @@ export function createStorage<T extends StorageValue>(
         ];
       }
       return allKeys
-        .filter((key) => filterKeyByDepth(key, opts.depth))
+        .filter((key) => filterKeyByDepth(key, opts.maxDepth))
         .filter(
           base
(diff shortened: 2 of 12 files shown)


src/drivers/netlify-blobs.ts
pi0, line 1: we should revert this

test/drivers/utils.ts
```diff
@@ -63,6 +63,29 @@ export function testDriver(opts: TestOptions) {
...
+    await ctx.storage.setItem("depth0:depth1:depth1_1", "test_data");
+    await ctx.storage.setItem("depth0:depth0_0", "test_data");
+    await ctx.storage.setItem("depth0:depth0_1", "test_data");
+    const depth0Keys = await ctx.storage.getKeys(undefined, { depth: 0 });
+    const depth1Keys = await ctx.storage.getKeys(undefined, { depth: 1 });
+    const depth2Keys = await ctx.storage.getKeys(undefined, { depth: 2 });
+
+    expect(depth0Keys).lengthOf(0);
+    expect(depth1Keys).includes.members(["depth0:depth0_0", "depth0:depth0_1"]);
+    expect(depth1Keys).not.includes.members([
```
pi0, line 77: We can merge these two tests with one explicit array check.

src/storage.ts
```diff
@@ -24,6 +24,22 @@ export interface CreateStorageOptions {
   driver?: Driver;
 }
 
+function filterKeyByDepth(key: string, depth: number | undefined): boolean {
```
pi0, line 27: Can be moved to `utils.ts`

src/storage.ts
```diff
@@ -24,6 +24,22 @@ export interface CreateStorageOptions {
...
+  if (depth === undefined) {
+    return true;
+  }
+
+  let substrCount = 0;
+  let index = key.indexOf(":");
+
+  while (index > -1) {
+    substrCount++;
+    index = key.indexOf(":", index + 1);
```
pi0, line 37: We might perform a quick bench on this, my guess is single `split` / count would be faster since reduces userland/native roundtrips

src/drivers/netlify-blobs.ts
43081j: typescript isn't happy if we do, since `ListOptions` doesn't satisfy `GetKeysOptions`

we have two options:

1. give `ListOptions` an index type (`[key: string]: any`) 
2. use `GetKeysOptions`

it got away with it in the past because the type it is implementing was just a `Record` afaict

src/storage.ts
43081j: last time we benchmarked this algo in a parser I was working on, the order was `indexOf > iterate-by-character > split`

split was pretty slow in comparison

src/drivers/netlify-blobs.ts
43081j: I went with `Record<string, any> & Omit<ListOptions, ...>` which satisfies the underlying type

src/storage.ts
pi0: Yep confirmed it is ~1.7x faster 👍🏼 

<summary>

```js
import assert from "node:assert";
import { summary, bench, run } from "mitata";

const impls = [
  {
    name: "indexOf (1)",
    fn: (key, depth) => {
      if (depth === undefined) {
        return true;
      }
      let substrCount = 0;
      let index = key.indexOf(":");
      while (index > -1) {
        substrCount++;
        index = key.indexOf(":", index + 1);
      }
      return substrCount <= depth;
    },
  },
  {
    name: "split",
    fn: (key, depth) => {
      return depth === undefined ? true : key.split(":").length <= depth + 1;
    },
  },
];

for (const impl of impls) {
  assert(impl.fn("a:b", 1) === true, `${impl.name} failed (true)`);
  assert(impl.fn("a:b:c", 1) === false, `${impl.name} failed (false)`);
}

summary(() => {
  for (const impl of impls) {
    bench(impl.name, () => {
      for (let i = 0; i < 1e5; i++) {
        impl.fn("a:b:c", 1);
        impl.fn("a:b", 1);
      }
    });
  }
});

await run();
```

</summary>`

43081j:
CI was failing because the file list wasn't sorted that we were asserting against

it may be worth us changing these tests in future to purge their storage after each test. it seems at the minute the tests can affect each other

test/drivers/utils.ts
```diff
@@ -63,6 +63,37 @@ export function testDriver(opts: TestOptions) {
...
+    await ctx.storage.setItem("depth0:depth0_0", "test_data");
+    await ctx.storage.setItem("depth0:depth0_1", "test_data");
+    const depth0Keys = await ctx.storage.getKeys(undefined, { depth: 0 });
+    const depth1Keys = await ctx.storage.getKeys(undefined, { depth: 1 });
+    const depth2Keys = await ctx.storage.getKeys(undefined, { depth: 2 });
+
+    depth1Keys.sort();
+    depth2Keys.sort();
+
+    expect(depth0Keys).lengthOf(0);
```
pi0, line 78: What would be the use case of `{ depth: 0 }` => [] you think?

IMO it could simply return top-level keys without `:`.

src/types.ts
```diff
@@ -18,6 +18,10 @@ export interface StorageMeta {
 // TODO: type ttl
 export type TransactionOptions = Record<string, any>;
 
+export type GetKeysOptions = TransactionOptions & {
+  depth?: number;
```
pi0, line 22: Thinking more, perhaps `maxDepth` is more readable?

src/storage.ts
```diff
@@ -362,11 +367,13 @@ export function createStorage<T extends StorageValue>(
...
       }
-      return base
-        ? allKeys.filter(
-            (key) => key.startsWith(base!) && key[key.length - 1] !== "$"
-          )
-        : allKeys.filter((key) => key[key.length - 1] !== "$");
+      return allKeys
+        .filter((key) => filterKeyByDepth(key, opts.depth))
+        .filter(
+          base
```
pi0, line 373: perf: I wish somehow we could merge filters. Current change adds one filter iteration even for users not setting `depth`

test/drivers/utils.ts
43081j: it would return keys which have no delimiter right now

are you saying you'd consider "top level" as the first delimited level? e.g. `foo:bar`, is top level because `foo` is the "root"?

src/types.ts
43081j: I agree, I already renamed the function parameter for the same reason

I'll update

test/drivers/utils.ts
pi0: Ah, then that's fine. We could perhaps add a `root` key to the test so that we test this behavior, it was a little misleading.

test/drivers/utils.ts
```diff
@@ -63,6 +63,37 @@ export function testDriver(opts: TestOptions) {
...
+    await ctx.storage.setItem("depth0:depth1:depth1_0", "test_data");
+    await ctx.storage.setItem("depth0:depth1:depth1_1", "test_data");
+    await ctx.storage.setItem("depth0:depth0_0", "test_data");
+    await ctx.storage.setItem("depth0:depth0_1", "test_data");
+    const depth0Keys = await ctx.storage.getKeys(undefined, { depth: 0 });
+    const depth1Keys = await ctx.storage.getKeys(undefined, { depth: 1 });
+    const depth2Keys = await ctx.storage.getKeys(undefined, { depth: 2 });
+
+    depth1Keys.sort();
+    depth2Keys.sort();
```
pi0, line 76: nitpick: we could inline it inside expect() to reduce lines :)

pi0 (approved):
Nice work ❤️

PR #548, opened 2024-12-28
test: clear storage after each test

43081j:
Clears the storage after each test such that they no longer rely on run order, etc.

Note that the two snapshot tests have been massively dumbed down. Though I'm not sure this is a problem since the underlying driver tests already test the data types etc. Leaving these to purely test that _a_ snapshot can be produced. 

let me know what you think


pi0 (approved):
This is much cleaner thnx ❤️

PR #551, opened 2024-12-29
feat(fs, fs-lite): native `maxDepth` support

43081j:
Adds native support for `maxDepth` to the `fs` and `fs-lite` drivers.

This works by traversing directories only until the depth has been met.

### DIFF
--- a/src/drivers/fs-lite.ts
+++ b/src/drivers/fs-lite.ts
@@ -73,8 +73,8 @@ export default defineDriver((opts: FSStorageOptions = {}) => {
       }
       return unlink(r(key));
     },
-    getKeys() {
-      return readdirRecursive(r("."), opts.ignore);
+    getKeys(_base, { maxDepth }) {
+      return readdirRecursive(r("."), opts.ignore, maxDepth);
     },
     async clear() {
       if (opts.readOnly || opts.noClear) {
--- a/src/drivers/fs-lite.ts
+++ b/src/drivers/fs-lite.ts
@@ -73,8 +73,8 @@ export default defineDriver((opts: FSStorageOptions = {}) => {
       }
       return unlink(r(key));
     },
-    getKeys(_base, { maxDepth }) {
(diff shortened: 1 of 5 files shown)


src/drivers/fs-lite.ts
```diff
@@ -73,8 +73,8 @@ export default defineDriver((opts: FSStorageOptions = {}) => {
       }
       return unlink(r(key));
     },
-    getKeys() {
-      return readdirRecursive(r("."), opts.ignore);
+    getKeys(_base, { maxDepth }) {
```
pi0, line 76: Do you mind refactoring it to pass whole `tOpts` to underlying utils? Reason: Avoid ref creation by destruct and also pass any future keys.

src/drivers/fs-lite.ts
```diff
@@ -73,8 +73,8 @@ export default defineDriver((opts: FSStorageOptions = {}) => {
...
+      return readdirRecursive(r("."), opts.ignore, maxDepth);
```
pi0, line 77: Style: can you please change it to avoid destructure? 

(generally, I keep this style to pass all keys to inner utils but I see `readdirRecursive` changed to be recursive so only a style change here for consistency)

test/drivers/fs.test.ts
```diff
@@ -40,6 +41,40 @@ describe("drivers: fs", () => {
...
+        await ctx.storage.setItem("depth-test/depth0/depth1/file2.md", "boop");
+        await ctx.storage.setItem("depth-test/depth0/depth1/file3.md", "boop");
+
+        const originalStat = await stat(
+          resolve(dir, "depth-test/depth0/file1.md")
+        );
+
+        expect(
+          (
+            await ctx.storage.getKeys(undefined, {
```
pi0, line 57: I think here we might use `ctx.driver.getKeys` to test native `Driver` functionality (without core fallback)

pi0 (approved):
🚀

PR #552, opened 2024-12-29
feat: add `flags` key to driver interface

### LINKED ISSUE #549: `capabilities` in drivers
context: #539

### Describe the feature

Just opening this to track how we end up implementing `capabilities` in drivers 

We need this so we can expose what a driver supports. For example, if a driver natively supports `maxDepth` when listing keys, we can avoid running the generic algorithm/filter 

Open questions: 

- currently, we conditionally enable some things based on presence of a method (e.g. raw methods). Do we want to continue with that, or introduce a capability? Where do we draw the line?

### Additional information

43081j:
resolves #549

Adds `flags` to the `Driver` type which can be used in future to determine if a driver supports a particular feature or requires some behaviour.

maybe something like this @pi0 ? and in a separate PR, add the flags to the fs drivers

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -23,7 +23,8 @@ export type GetKeysOptions = TransactionOptions & {
 };
 
 export interface DriverFlags {
-  supportsMaxDepth?: boolean;
+  maxDepth?: boolean;
+  ttl?: boolean;
 }
 
 export interface Driver<OptionsT = any, InstanceT = any> {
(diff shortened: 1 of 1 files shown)


src/types.ts
```diff
@@ -22,8 +22,13 @@ export type GetKeysOptions = TransactionOptions & {
   maxDepth?: number;
 };
 
+export interface DriverFlags {
+  supportsMaxDepth?: boolean;
```
pi0, line 26: I assume (most of the) options will be "capabilities" which imply `supports` part. It can reduce verbosity and few bytes in bundles to omit.

```suggestion
  maxDepth?: boolean;
  ttl?: boolean;
```

pi0 (approved):

PR #554, opened 2024-12-29
feat: strongly type extended drivers

43081j:
Currently, a driver can define its own method signatures but they will be lost thanks to `DriverFactory` returning a `Driver` rather than the original type.

This means things like the netlify driver never actually exposed their extended types (e.g. special options in `getKeys`).

```ts
import driver from "./netlify-blobs";

const instance = driver();

// BEFORE

instance.getKeys;
//`        ^? getKeys(base?: string, opts: GetKeysOptions)

// AFTER

instance.getKeys;
//         ^? getKeys(base?: string, tops?: GetKeysOptions & Omit<ListOptions, "prefix" | "paginate">)
```

With these changes, a driver can implement `Driver` but add its own types on top which are then exposed.

### DIFF
--- a/src/drivers/session-storage.ts
+++ b/src/drivers/session-storage.ts
@@ -7,10 +7,10 @@ const DRIVER_NAME = "session-storage";
 
 export default defineDriver((opts: SessionStorageOptions = {}) => {
   return {
-    name: DRIVER_NAME,
     ...localstorage({
       windowKey: "sessionStorage",
       ...opts,
     }),
+    name: DRIVER_NAME,
   };
 });
--- a/src/drivers/capacitor-preferences.ts
+++ b/src/drivers/capacitor-preferences.ts
@@ -8,46 +8,44 @@ export interface CapacitorPreferencesOptions {
   base?: string;
 }
 
-export default defineDriver<CapacitorPreferencesOptions, typeof Preferences>(
-  (opts) => {
-    const base = normalizeKey(opts?.base || "");
-    const resolveKey = (key: string) => joinKeys(base, key);
+export default defineDriver((opts: CapacitorPreferencesOptions) => {
+  const base = normalizeKey(opts?.base || "");
+  const resolveKey = (key: string) => joinKeys(base, key);
 
-    return {
-      name: DRIVER_NAME,
-      options: opts,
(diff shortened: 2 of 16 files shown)


43081j:
I think we should do this for the repo in its current state

But given what you said in #555 , it made me wonder about what the right interface is

I wonder if in v2, we should actually have a fixed interface (like before this change). And an option which contains driver specific options, instead of merging the two

```ts
getItem("key", {
  driverOptions: {
    somethingDriverSpecific: true
  }
})
```

src/drivers/session-storage.ts
```diff
@@ -7,10 +7,10 @@ const DRIVER_NAME = "session-storage";
 
 export default defineDriver((opts: SessionStorageOptions = {}) => {
   return {
-    name: DRIVER_NAME,
     ...localstorage({
       windowKey: "sessionStorage",
       ...opts,
     }),
+    name: DRIVER_NAME,
```
pi0, line 14: Reminder this another valid fix

pi0:
Appreciate your work on this PR ❤️

`defineDriver<T>(): Driver` is intentional at least for built-in drivers so we only assume the "common" interface only also explicit generic `<T>` keeps types being isolated.

In v2, it shall be a global namespace interface with explicit keys for each driver (which is, unfortunately, a breaking change for v1 as we used top-level mixed keys)

```js
{
  ttl: 1200, // Generic
  http: { headers: { } } // driver specific
}
```

43081j:
The netlify driver specifies it's own types which currently have no use, since we only expose a `Driver`

Was that intentional too? That's mostly why I did this change

pi0:
Indeed, they are only usable internally now for those drivers.

Also consider, that passing driver-driver options to each transaction is something more of an escape hatch / advanced purpose. The key point of using unstorage abstraction is that an operation works regardless of the underlying driver.

43081j:
I see, it was just there to help with the internal types 

And yes, I would much rather it return a `Driver` so there is one interface regardless of driver 

Then if we must, have an option in there that is passed through to the underlying driver. Instead of changing the method signature

pi0:
(lets discuss it later)

PR #555, opened 2024-12-29
feat: pass lru options through

### LINKED ISSUE #466: Support get, set method opts parameters
### Describe the feature

I found that the opts parameter exists, but the specific driver has not implemented it. Is there a plan to support it?
https://github.com/unjs/unstorage/blob/bedbb3dd367735d565134283fccff7d83f4df883/src/storage.ts#L217

### Additional information

43081j:
Fixes #466

This changes the `lru-cache` such that the following methods accept LRU options (and pass them through to the underlying methods):

- `setItem`
- `setItemRaw`
- `hasItem`

### DIFF
--- a/src/drivers/lru-cache.ts
+++ b/src/drivers/lru-cache.ts
@@ -1,5 +1,6 @@
 import { defineDriver } from "./utils";
 import { LRUCache } from "lru-cache";
+import { type TransactionOptions, type StorageValue } from "../types";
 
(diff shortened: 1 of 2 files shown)


pi0:
I appreciate the PR but I want to wait for v2 to standardize opts namespace before exposing more possible options (otherwise we introduce a feature that will be shortly broken) -- PS: we might want to limit some of options for lru.

PR #556, opened 2024-12-29
fix(prefixStorage): wrap shorthand aliases

CorentinTh:
When creating a prefixed storage, the `get`, `set`, `has`, `del`, `remove` aliases don't work as intended (they do not get the defined prefixe) as they are not defined in the `storageKeyProperties` in https://github.com/unjs/unstorage/blob/main/src/utils.ts

- Added `get`, `set`, `has`, `del`, `remove` in `storageKeyProperties`
- Added regression tests 

Related with breaking changes : https://github.com/unjs/unstorage/pull/455


pi0 (approved):
Thnx ❤️

Issue #528, opened 2024-12-18
`storage.get(key, { type })`


pi0:
### Describe the feature

Currently, unstorage has two ways to access items:

- `getItem(key)`: Returns a json parsed value if parsable or fallback to text
- `getItemRaw(key): Returns raw storage value (Typically arrayBuffer but unpredictable)

This is an unpredictable behavior and also introduced two sets of methods. 

We could support `storage.get(key, { type: "text" | "json" | "bytes" | "stream" })` API to unify this and also make behavior more predictable.

```ts
await storage.get("key") // Defaults to json > text as before using destr
await storage.get("key", { type: "text" )  // Avoid serialization
await storage.get("key", { type: "json" )  // Strictly use JSON.parse
await storage.get("key", { type: "bytes" ) // Return Buffer or Unit8Array
await storage.get("key", { type: "bytes" ) // Return a readable stream
```

[undio](https://github.com/unjs/undio) can potentially be integrated or we might end-up directly adding small utils in unstorage core to convert

### Additional information

- [X] Would you be willing to help implement this feature?

cjpearson:
This looks like a nice improvement to me. I suppose the typing would also be improved by looking at the `type` option? Does it make sense to also include blob as a type?

lishaduck:
Instead of using `type`, what about integrating with [standard-schema](https://github.com/standard-schema/standard-schema)?  That'd let you use zod, valibot, etc, parsers without a) pulling them in, b) picking a specific one, but allows users to do complex validations.

pi0:
Validation is another topic (for object/json), these are data types.

sandros94:
Oh, I didn't notice this!

What is the current plan for the actual data type within the storage itself? Personally I think this should be customizable depending on the actual project's requirement.

sandros94:
Also, could this be the correct time to add optional compression?

Recently, as I was experimenting with custom drivers with built-in compression, I've noticed that I could directly return the compressed data (1:1 from the storage) as the response from Nitro event handlers (while also adding correct headers). This became particularly useful for me because I could cache event handlers and Nuxt pages more efficiently (about 2/3 of server memory saved. No benchmark yet, just watching the resource monitor every few hours)

cjpearson:
I don't know if you've already seen it, but there is an open issue for compression: https://github.com/unjs/unstorage/issues/94 

I've also been using a custom driver to implement compression for unstorage and using it as the backing store for Nitro's page cache.

[...]

PR #561, opened 2025-01-02
feat(redis): add `saveRawAsBinary` driver option

### LINKED ISSUE #559: [redis] Unnecessary base64 encoding for binary data
### Describe the feature

Currently the redis driver does not implement `getItemRaw`/`setItemRaw`. It uses the default method of `serializeRawData` + `setItem`.

Since ioredis can binary data, it should be unnecessary to encode it as base64. Directly setting a `Buffer` would have less overhead and be more space-efficient. However, there would be compatibility issues with existing data. Perhaps it could be implemented as an option or separate driver?

https://github.com/redis/ioredis#handle-binary-data

Here's an example implementation. It would still need to be updated to handle non-Buffer values.

```ts
async getItemRaw(key) {
  const value = await getRedisClient().getBuffer(p(key));
  return value ?? null;
},
async setItemRaw(key, value, tOptions) {
  if (value instanceof Uint8Array) {
    value = Buffer.from(value, value.byteOffset, value.byteLength)
  }
  const ttl = tOptions?.ttl ?? opts.ttl;
  if (ttl) {
    await getRedisClient().setBuffer(p(key), value, "EX", ttl);
  } else {
    await getRedisClient().setBuffer(p(key), value);
  }
},
```

This could perhaps be done with https://github.com/unjs/unstorage/issues/528 and I think that would avoid the breaking aspects. For example `setItemRaw` could continue to base64 encode and save as a string, but `setItem(k, v, { type: 'bytes'})`, could save as binary.

### Additional information

cjpearson:
Resolves #559 

This change adds a new `saveRawAsBinary` option to the redis driver. When enabled, the driver will use the `setBuffer` and `getBuffer` methods of `ioredis` to directly save `Buffer`s rather than base64 encoding and saving as a string. The option is not enabled by default as that would be a backwards incompatible change. 

When #528 is implemented, this could be made the default for `{type: 'bytes'}`. (Perhaps then `saveBytesAsBinary` is a better name?) It would also be possible to drop the base64 encoding entirely and instead allow users to manually encode and save as `{ type: 'text' }`.


src/_utils.ts
pi0, line 1: this is an unrelavant change please revert.

src/drivers/redis.ts
```diff
@@ -75,6 +87,42 @@ export default defineDriver((opts: RedisOptions) => {
         await getRedisClient().set(p(key), value);
       }
     },
+    async getItemRaw(key) {
+      if (opts.saveRawAsBinary) {
```
pi0, line 91: I think in next major we should unconditionally do this.  We can optionally add `getItemRaw` and `setItemRaw` keys in current version. 

(if it is too complex description, please simplify your changes to have `saveRawAsBinary` behavior without flag I can push other reactors to finalize)

cjpearson: > We can optionally add getItemRaw and setItemRaw keys in current version. (as a result, driver doesn't need to directly do fallback base64 handling, core does)

This approach sounds much better to me, I'll try it out.

> I think in next major we should unconditionally do this. 

I would also favor this. My only concern is that it would require user action to make existing data accessible. But with clear enough migration notes it's probably ok. Perhaps a dev warning could also be added in a future minor release?

pi0: Unstorage (core) can also be smart for raw operations returning `string` starting with a signature to reverse it so we can remain backward compatible.

BTW also consider `raw` methods are still marked as experimental (and that's partially why for nitro cache I didn't adopt raw to minimize affected users for such changes).

cjpearson: I've updated the PR to only add `getItemRaw`/`setItemRaw` when the option is enabled and to add some tests for the behavior. 

I think we need to know upfront whether to expect binary or string data because that determines whether we call `ioredis.get` or `ioredis.getBuffer`. There may still be a way to catch a mismatch, but it's a good point that raw is still experimental, and there is a migration path.

src/drivers/redis.ts
```diff
@@ -56,25 +63,67 @@ export default defineDriver((opts: RedisOptions) => {
...
+          value,
+          value.byteOffset,
+          value.byteLength
+        );
+      }
+    } else {
+      throw createError(DRIVER_NAME, "Expected Buffer or Uint8Array");
+    }
+
+    await setItem(key, valueToSave, tOptions);
```
pi0, line 109: For fallback, while core ensures `values` is string, here we cannot assume the same

docs/2.drivers/redis.md
```diff
@@ -72,7 +72,7 @@ const storage = createStorage({
 - `cluster`: List of redis nodes to use for cluster mode. Takes precedence over `url` and `host` options.
 - `clusterOptions`: Options to use for cluster mode.
 - `ttl`: Default TTL for all items in **seconds**.
-- `useRaw`: If enabled, `getItemRaw` and `setItemRaw` will use binary data instead of base64 encoded strings. (this option will be enabled by default in the next major version.)
+- `raw`: If enabled, `getItemRaw` and `setItemRaw` will use binary data instead of base64 encoded strings. (this option will be enabled by default in the next major version.)
```
cjpearson, line 75: I would suggest something like `binary`/`useBinary` here. I think it makes the effect a bit more clear because it changes how raw data is stored, but does not enable/disable the ability to save raw data. Or it could be inverted to `base64` so in the future the default value can be false.

pi0: I agree that in general `binary` could be more describing than `raw`. In the context of unstorage, we call binary-compatible feature `raw` which was the reason I was thinking to use this. 

Considering it is for short-term solution, are you happy we go with this?

(Also check fb2977a9855bad1637798d87eb82ad850bc87e70, I have extended normalization)

cjpearson: Extending normalization to additional variants of TypedArray makes sense, but it should only allow for binary data and continue to throw when other data is passed. Otherwise it would lead to some unexpected behavior as you save an object but get a Buffer back. 

That's also why I think the different naming is helpful. `binary` means that `setItemRaw` accepts binary and `getItemRaw` returns binary. Whereas the existing `raw` behavior accepts binary and other things.

pi0: This is consistent with all other drivers (both those support binary and those not) -- happy to change it in the future bu† normally, users should use same `key` either for raw or non raw purposes.

cjpearson: Hm would it be possible to add an additional option in that case? Either on the driver or the method? What I'm looking for is a way to save binary data and ensure it is binary and saved as binary and when getting, that I get a Buffer in return. 

My assumption is that is how `getItem(key, { type: 'bytes'})`/`setItem(key, value, { type: 'bytes'})` would work in the future.

pi0: Yes exactly for `type: 'bytes'` we can get strict.

Currently, if someone passes a string to `setItemRaw`, `getItemRaw` gives a `Buffer` representation of `Utf8` bytes which isn't that odd ,is it?

cjpearson: Yeah that's what I find a bit unexpected 😅 

Do you think there would be an issue if we added `type: 'bytes'` as a method option here? It would be just for the redis driver at the moment, but done in a forward compatible manner.

Or is it better to use something else and save that wording for later

PR #565, opened 2025-01-04
feat(azure-storage-blob): add raw support

peterbud:
This PR adds raw item get/set support, similar to abandoned #442, but also adds test for that new procedures.

Also I have noticed that the current `removeItem` test fails, as `delete` operation with non-existent keys leads to an error. I _assume_ the intention is not to fail in such cases, so I have not changed the `removeItem` test, but added a check before attempting to delete the blob at the Azure Blob driver.

### DIFF
--- a/src/drivers/azure-storage-blob.ts
+++ b/src/drivers/azure-storage-blob.ts
@@ -10,10 +10,10 @@ export interface AzureStorageBlobOptions {
   /**
    * The name of the Azure Storage account.
    */
-  accountName: string;
+  accountName?: string;
 
   /**
-   * The name of the storage container. All entities will be stored in the same container.
+   * The name of the storage container. All entities will be stored in the same container. Will be created if it doesn't exist.
    * @default "unstorage"
    */
   containerName?: string;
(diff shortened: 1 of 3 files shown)


itpropro:
Thanks for bringing this up again, I unfortunately missed the PR.
Let me go over the tests to validate we are fine here so we can proceed as soon as possible.

peterbud:
Hi @itpropro , is there any update on your review please?

itpropro:
> Hi @itpropro , is there any update on your review please?

Hey,
I added some additions to improve test coverage. Thanks for also fixing the removeItem inconsistency. Can you explain what the  "properly encodes raw items" test covers that is not already covered by the "raw support" test in the test suite?
If it covers the same as the internal test, I don't think it's necessary.
After that we are good to go from my perspective, I tested the current code against Azurite and an online instance of a storage account.

peterbud:
> Can you explain what the "properly encodes raw items" test covers that is not already covered by the "raw support" test in the test suite?

I did not know that there is a default raw support test. I have seen that the fs driver has a [raw items test](https://github.com/unjs/unstorage/blob/main/test/server.test.ts#L72), and I assumed this had to be added for drivers which implement these additional methods.

If not needed, then let's remove this test!

itpropro:
No problem, all drivers use the `testDriver` test suite as a baseline. That one has a rawItem test here https://github.com/unjs/unstorage/blob/67b09a164fc9734435b29d83ccabcc1b0220389b/test/drivers/utils.ts#L136.

itpropro:
I removed the redundant test. Should I extract the additional added tests to a separate PR @pi0 ?

src/drivers/azure-storage-blob.ts
```diff
@@ -24,41 +24,69 @@ export interface AzureStorageBlobOptions {
...
 export default defineDriver((opts: AzureStorageBlobOptions) => {
   let containerClient: ContainerClient;
+  const endpointSuffix = opts.endpointSuffix || ".blob.core.windows.net";
   const getContainerClient = () => {
     if (containerClient) {
       return containerClient;
     }
-    if (!opts.accountName) {
-      throw createError(DRIVER_NAME, "accountName");
+    if (!(opts.connectionString || opts.sasUrl) && !opts.accountName) {
```
43081j, line 57: this negated logic reads strangely

`accountName` only becomes optional if you provide a `sasUrl` or a `connectionString` right?

this could be made clearer by something like this:

```ts
if (!opts.sasUrl && !opts.connectionString && !opts.accountName) {
```

src/drivers/azure-storage-blob.ts
```diff
@@ -100,13 +129,40 @@ export default defineDriver((opts: AzureStorageBlobOptions) => {
...
         .upload(value, Buffer.byteLength(value));
     },
+    async setItemRaw(key, value) {
+      await getContainerClient()
+        .getBlockBlobClient(key)
+        .upload(value, Buffer.byteLength(value));
+    },
     async removeItem(key) {
-      await getContainerClient().getBlockBlobClient(key).delete();
+      const exists = await getContainerClient()
```
43081j, line 158: you can actually call `deleteIfExists({ deleteSnapshots: "include" })` iirc

src/drivers/azure-storage-blob.ts
peterbud: I have changed it

src/drivers/azure-storage-blob.ts
peterbud: Great suggestion. Done

peterbud:
Hi team, is there anything else to be done on this PR to get it merged?

src/drivers/azure-storage-blob.ts
```diff
@@ -24,41 +24,69 @@ export interface AzureStorageBlobOptions {
...
-      throw createError(DRIVER_NAME, "accountName");
+    if (!opts.connectionString && !opts.sasUrl && !opts.accountName) {
+      throw createError(DRIVER_NAME, "missing accountName");
     }
     let serviceClient: BlobServiceClient;
     if (opts.accountKey) {
       // StorageSharedKeyCredential is only available in Node.js runtime, not in browsers
       const credential = new StorageSharedKeyCredential(
-        opts.accountName,
+        opts.accountName!,
```
43081j, line 64: do you need this non-null assertion anymore?

```suggestion
         opts.accountName,
```

43081j:
it all looks good to me

i left one minor comment. we need Pooya to review it now and merge it

pi0 (approved):
❤️

Thanks for PR. I trust @43081j made full review

src/drivers/azure-storage-blob.ts
itpropro: This is from my initial implementation. I tried to incorporate as many combinations as possible, as not all of the combinations are mutually exclusive.

Issue #581, opened 2025-01-24
Configure from URL.


AntoineMontane:
### Describe the feature

Would be glad to have single variable configuration as we have in some db library (sqlite:///, postgres:///, etc.), that would ease per-environment configuration.

Some drivers already have natural url scheme like file:// http://, s3://, etc that we could handle.

I envision an async createStorageFromUrl(url: string) that would load dynamically the driver associated to the url, and then let each driver handle it's own url scheme.

I would implement for a given set of driver and let more knowledgeable person handle the scheme for other drivers.

### Additional information

- [x] Would you be willing to help implement this feature?

PR #582, opened 2025-01-25
Feature URI : load driver by uri like file:/// http:/// etc

AntoineMontane:
solve #581


pi0:
This is a nice idea however as it has bundle size implications, I would keep it completely separate.

Can you please as first step, deduce diff Generate a new entry in dist that is an object mapping from URI scheme like `"file"`, to driver object. (there should be no changes in driver src files)

src/drivers/fs-lite.ts
```diff
@@ -20,7 +26,19 @@ const PATH_TRAVERSE_RE = /\.\.:|\.\.$/;
 
 const DRIVER_NAME = "fs-lite";
 
-export default defineDriver((opts: FSStorageOptions = {}) => {
+const defineDriver = defineDriverWithUriHandler((uri: string) =>
+  /^file:.*lite=true/.test(uri)
```
pi0, line 30: Please revert this changes from `src/drivers/*` (they should be unchanged) -- we need to define a generic map from uri to dynamic imports in a new file in `src/.ts`

test/drivers/fs.test.ts
```diff
@@ -65,4 +66,17 @@ describe("drivers: fs", () => {
       });
     },
   });
+
+  it("fromUri", async () => {
```
pi0, line 70: Please revert per-driver additional tests. Generic dynamic loader should be tested in a new `tests/.ts` entry

AntoineMontane: Some tests are generic, yes, but how the uri is processed is driver dependent.
In per-driver tests i checks that the driver handles the designed uri scheme + process correctly uri parameters.

AntoineMontane:
I think i should go back to the previous approach to export the regexp from each driver module. And keep the new one for the loader. What do you think of this ?

src/drivers/fs-lite.ts
pi0: Yes i want exactly one dedicated file that has the map. We don’t need regex (or dedocators), protocol is first part of the URI, and should be key in this map object.

Contents are drivers (including defineDriver) should completely be part of (lazy) dynamic import.

AntoineMontane: Ok. Let say drivers exports a protocol string, these strings are gathered with gen-drivers into a map (drivername -> protocol) in _drivers.ts
A loader.ts method uses the map to find the correct driver name for a given URI.
I would leave the dynamic import into a separate module : it's a feature that could be usefull or not depending on the project.

test/drivers/fs.test.ts
AntoineMontane: how a generic conversion would work ?
we could do `protocol://?k1=v1&k2=v2... =>  opts {k1: v1, k2: v2}`

But i would expect more natural uri, following usual protocol like this : 

```
http://host/path?Authorization=key
=> opts { base: "http://host/path",  header: Authorization: "key" }
```

```
github://[apiUrl]/org/repo/branch 
=> ops { repo: "org/repo", branch: "branch", apiUrl?: apiUrl }
```

Thats why i would have gone with protocol specific knowledge at the driver level.

What do you think of this ?

pi0: `[driver]://[base]?[opts]`

src/loader/all.ts
```diff
@@ -0,0 +1,69 @@
...
+import nullDriver from "unstorage/drivers/null";
+import overlay from "unstorage/drivers/overlay";
+import planetscale from "unstorage/drivers/planetscale";
+import redis from "unstorage/drivers/redis";
+import s3 from "unstorage/drivers/s3";
+import sessionStorage from "unstorage/drivers/session-storage";
+import uploadthing from "unstorage/drivers/uploadthing";
+import upstash from "unstorage/drivers/upstash";
+import vercelBlob from "unstorage/drivers/vercel-blob";
+import vercelKv from "unstorage/drivers/vercel-kv";
```
pi0, line 32: We should generate dynamic import for drivers.

src/loader/index.ts
```diff
@@ -0,0 +1,35 @@
+import { parseQuery } from "ufo";
+import { createError } from "../drivers/utils";
+import { coerceQuery } from "./utils";
+import type { Driver, DriverFactory } from "../types";
+
+const RE = /^(?<protocol>[^:?]+)(:(?<base>[^?]*))?(\?(?<query>.*))?$/;
```
pi0, line 6: We should use `new URL` as parser.

test/loader.test.ts
```diff
@@ -0,0 +1,40 @@
...
+  options: options,
+  hasItem: () => false,
+  getItem: () => null,
+  getKeys: () => [],
+}));
+
+describe("loader", () => {
+  it("load driver", async () => {
+    const driver = loadFromUri(
+      'test:abc?string=def&number=1&boolean=true&array=[2,3,4]&object={"h":5,"i":6,"j":"7"}',
```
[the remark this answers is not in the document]
AntoineMontane: not necessarily : we want to still be able to express relative uri not starting with /. Ex with http

`http:relative/path => base: "relative/path`

`http:/absolute/path => base: "/absolute/path`

`http://fqdn/path => base: "http://fqdn/path`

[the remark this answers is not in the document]
pi0: supporting relative urls for http (that only works in the browser) is really an edge case in unstorage. It can be supported still as `'http:/.?base=./relative'`

src/loader/index.ts
AntoineMontane: i'm not convinced by the ability of new URLto express relative urls:

```
> new URL('file:abc/def')
URL {
  href: 'file:///abc/def',
  origin: 'null',
  protocol: 'file:',
  username: '',
  password: '',
  host: '',
  hostname: '',
  port: '',
  pathname: '/abc/def',
  search: '',
  searchParams: URLSearchParams {},
  hash: ''
}
```
```
> new URL('file:/abc/def')
URL {
  href: 'file:///abc/def',
  origin: 'null',
  protocol: 'file:',
  username: '',
  password: '',
  host: '',
  hostname: '',
  port: '',
  pathname: '/abc/def',
  search: '',
  searchParams: URLSearchParams {},
  hash: ''
}
```
```
> new URL('http:/abc/def')
URL {
  href: 'http://abc/def',
  origin: 'http://abc',
  protocol: 'http:',
  username: '',
  password: '',
  host: 'abc',
  hostname: 'abc',
  port: '',
  pathname: '/def',
  search: '',
  searchParams: URLSearchParams {},
  hash: ''
}
```

pi0:
Unstorage is primarily targeting server runtimes. Supporting relative URLs is really an edge case that possible in browser-only environments.

(and we should have **one** entry exporting dynamic imports inside it / no extra browser.ts)

AntoineMontane:
Is it wrong to support browser runtimes ? In fact i started my development especially for this cross-environment ability.
Relative url for file: is not an edge case.

src/loader/index.ts
AntoineMontane: i have tries with ufo.parseUrl without much success either

AntoineMontane: url-parse library looks to be ok ; yet, i'm not sure it's a good idea to bring in another url parsing library

pi0:
js runtime support depends on drivers (if they support broeser they do) — we shouldn’t make an exception entey for browser.

re parsing, it kinda makes sense for some drivers but also incompatible with URL standards. To be honest i really have to locally play with some ideas my self at this point.

Please ping me when your work finished later when i could will try to work on it.

AntoineMontane:
Is it possible someway to mark node:fs external so that it doesn't represent an issue at client bundle build ?

pi0:
that's a bundler-specific config people should do.

Also, we need to have **exactly** syntax below in the generated list:

```js
() => import("unstorage/drivers/name")
```

this is because when using `import("unstorage/drivers/${name}")` bundler cannot statically analyze and has to scan all possible patterns.

AntoineMontane:
I've
* refactored uri to url,
* refactored the factoryloader to import with raw string
* dropped the browser bundle

It would be less deceptive to only include in the all bundle drivers that were verified or reworked to correctly handle the url protocol.

AntoineMontane:
Hey @pi0 Do you have a good advice to handle driver import location in the all bundle ? 
I don't want to import from `unstorage/drivers/` because it's not available until build and add a step before testing.
If i import relatively from `../drivers/`, it works in test but fails in production as the path is different (`../../drivers/`)

src/drivers/http.ts
```diff
@@ -5,16 +5,22 @@ import { joinURL } from "ufo";
...
+  const base =
+    opts.base.startsWith("//") && opts.scheme
+      ? `${opts.scheme}:${opts.base}`
+      : opts.base;
+
+  const r = (key: string = "") => joinURL(base!, key.replace(/:/g, "/"));
 
   const rBase = (key: string = "") =>
-    joinURL(opts.base!, (key || "/").replace(/:/g, "/"), ":");
+    joinURL(base!, (key || "/").replace(/:/g, "/"), ":");
```
pi0, line 23: Please split this into a new PR with a clear description of new behavior

src/drivers/s3.ts
```diff
@@ -36,7 +48,7 @@ export interface S3DriverOptions {
   /**
    * The name of the bucket.
    */
-  bucket: string;
+  bucket?: string;
```
pi0, line 51: `endpoint` should not be deprecated because of a new feature. If you like to introduce `sessionToken` please open one PR for it.

If you want to introduce `base`, another PR.

src/loader/all.ts
pi0, line 1: (TODO for myself): this should be auto generated

PR #584, opened 2025-01-29
feat(azure-drivers): add managed identity client ID support

itpropro:
Enhance Azure drivers to support user-assigned managed identity authentication by adding an optional `managedIdentityClientId` parameter. This allows choosing the correct user assigned managed identity to use.

Changes:
- Added `managedIdentityClientId` option to driver interfaces to use with DefaultAzureCredential if provided
- Improved authentication flexibility for Azure services
- Fixing a bug with CosmosDB syntax that changed that didn't came up due to disabled test suites
- Fixed `removeItem` throws in CosmosDB driver
- Improved stability of Key Vault and it's test suite by introducing a wait of 750ms (average time to purge) only after `clean` if purge is provided
- Added a parameter to disablePurging for Key Vaults with enabled purge protection

All changes are tested against the local emulator (if available) and live Azure resources.

### DIFF
--- a/src/drivers/azure-app-configuration.ts
+++ b/src/drivers/azure-app-configuration.ts
@@ -4,6 +4,12 @@ import { DefaultAzureCredential } from "@azure/identity";
 
 export interface AzureAppConfigurationOptions {
   /**
+   * Optional client ID for managed identity. Needed if using one or more user assigned managed identity to authenticate.
+   * @default null
+   */
+  managedIdentityClientId?: string;
+
+  /**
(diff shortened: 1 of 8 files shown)


itpropro:
> Appreciate the PR but please consider splitting changes next times, it is clearly written in PR template.
> 
> > 2. Please keep your changes minimal and split them if you need to.

I can of course split this, the errors just occured "on accident" when running the test suites and for the changes to be validated properly, I needed the test suites to be representative.
It's no problem to split out the the fixes separately if you want, just wanted to save you the review work :)

pi0:
no worries on this one is good 👍🏼  (sorry might be back on this with some delay)

PR #586, opened 2025-01-29
refactor(redis): remove colon from base prefix

### LINKED ISSUE #585: Deprefixing doesn't remove colon in Redis Driver
### Environment

Node.js v23.6.1

### Reproduction

Call `getKeys` with `base` being set up

### Describe the bug

Calling getKeys returns wrong keys(with `:` at the begging of the keys)

### Additional context

_No response_

### Logs

```sh

```

denisugo:
Resolves #585


43081j:
out of interest, how did you bump into this? do you use the driver directly? i.e. rather than `Storage`

i think your fix makes sense, but `getKeys` in the result of `createStorage(...)` already strips `:` from the start of these keys

so would be good to know what situation you were in where you got the underlying key, rather than the one `Storage` has already normalised

denisugo:
I copied the code for the redis driver and used it as a **custom driver**, however the implementation of `d()` and `p()` remained unchanged(it was simply copy-pasted) and it didn't support the colon removal.

43081j:
so you used it directly rather than via `Storage`/`createStorage`?

good to know if so!

basically if you make it via `createStorage`, it will normalise those keys for you and strip the `:` already. 

though i still think its a good idea for the driver itself to do that 👍

denisugo:
Well, I did use `createStorage`, but I also used a custom redis driver, whose entire implementation was copied from the standard redis driver used by unstorage

denisugo:
So, the code looks like

```
createStorage({driver: myRedisDriver(...)})
```

pi0 (approved):
Thanks!

PR #591, opened 2025-02-06
fix string stringify

### LINKED ISSUE #590: store strings like '83528e0696' result in `Infinity`
### Environment

unstorage 1.14.4

### Reproduction

https://codesandbox.io/p/sandbox/bold-waterfall-qgqgpg

```js
import { createStorage } from "unstorage";

const storage = createStorage();

storage
  .setItem("foo", "83528e0696")
  .then(() => {
    return storage.getItem("foo");
  })
  .then((result) => {
    console.log(result);
  });
```

### Describe the bug

The console logs `Infinity`.
Expected to be '83528e0696'

### Additional context

_No response_

### Logs

```sh

```

intellild:
resolves #590

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -1,3 +1,4 @@
+import { isPrimitive, isPureObject } from "./_utils.ts";
 import type { Storage, StorageValue } from "./types";
 
 type StorageKeys = Array<keyof Storage>;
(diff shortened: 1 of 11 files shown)


src/utils.ts
```diff
@@ -98,3 +99,23 @@ export function filterKeyByBase(
 
   return key[key.length - 1] !== "$";
 }
+
+export function stringify(value: any): string {
+  if (typeof value === "string") {
+    return JSON.stringify(value);
```
43081j, line 105: this would be a breaking change i think

since it means people who were previously storing strings as-is, will now store them as their JSON equivalents (`"foo"` instead of `foo`)

43081j:
ultimately this is because we `stringify(value)` before we set it, then we basically `JSON.parse` it on the way out (`getItem`)

that means `setItem(key, 100)` and `setItem(key, '100')` behave the same currently (in `main`). so you've uncovered that we're actually being lossy - in that if we `getItem` what was originally a string, we can never get back to the string (i.e. we don't know it was ever a string, we've lost that information)

i wonder if we should just be storing everything as the JSON representation... we will need @pi0 for this one.

pi0:
Thanks for working on PR @intellild. As @43081j pointed out, this is a breaking change. We could do it for unstorage v2 while considering how (data) migration happens.

In the meantime, this change is not acceptable. We might think of another way like an option for `setItem(key, value, { preserveTypes: true })` to preserve string types with `""`.

Lets' keep track on #277

PR #592, opened 2025-02-12
fix(s3): missing `getInstance`

sandros94:
Missing `getInstance` for the `S3` driver, this should allow to presign urls based on the existing driver options.

### DIFF
--- a/src/drivers/s3.ts
+++ b/src/drivers/s3.ts
@@ -162,6 +162,7 @@ export default defineDriver((options: S3DriverOptions) => {
   return {
     name: DRIVER_NAME,
     options,
+    getInstance: getAwsClient,
     getItem(key) {
       return getObject(key).then((res) => (res ? res.text() : null));
     },


43081j:
can you add a test too?

similar to the one in the redis tests:

```ts
      it("exposes instance", () => {
        expect(driver.getInstance?.()).toBeInstanceOf(ioredis.default);
      });
```

test/drivers/s3.test.ts
```diff
@@ -12,15 +12,16 @@ const region = process.env.VITE_S3_REGION;
 describe.skipIf(
   !accessKeyId || !secretAccessKey || !bucket || !endpoint || !region
 )("drivers: s3", () => {
+  const driver = s3Driver({
```
43081j, line 15: for whatever reason, vitest does actually call this overall callback even if the `skipIf` is true (i.e. if it is skipped)

so this still needs to be:

```suggestion
  const driver = () => s3Driver({
```

since, if `endpoint` isn't set, it'll throw even if this test is skipped

sandros94:
I'm noticing that the `can access directly with / separator` test is failing on my end, but it is probably caused by the fact I'm using a _mostly-compatible_ S3 I have available

pi0:
I'm also thinking if it is too early to expose a "client" off s3. If you check aws4fetch implementation, we mainly use it for signing. Many of other functionalities are internal.

Changing the return type after we expose it would be a breaking change also..

We could perhaps return an object like `{ sign }` to expose limited functionality but explicit interface so if later we change internal library it won't be a breaking change.

sandros94:
> Changing the return type after we expose it would be a breaking change also..
> 
> We could perhaps return an object like `{ getURL }` to expose limited functionality but explicit interface so if later we change internal library it won't be a breaking change.

Oh, indeed.

I'll study this and propose a few potentially useful utils. One for sure being `getUrl` (for pre-signed urls) as it will be a useful one IMO

PR #593, opened 2025-02-14
docs(redis): update hashtags note

cjpearson:
This change mainly corrects the spelling of "hashtags" and adds links to the redis docs. (They also use "hash tags" sometimes though, so I'm not sure which is best.) 

But I think it might also be helpful to add a bit of nuance to this section. Forcing every item to the same slot will avoid the CROSSSLOT error, but isn't generally recommended. My assumption is a strategy similar to the one shown in the redis docs (slotting based on the user ID) would also work with unstorage, but there may be limitations such as `getKeys` not working. 
What are your thoughts here? (I can split this into a separate issue if it helps)

### DIFF
--- a/docs/2.drivers/redis.md
+++ b/docs/2.drivers/redis.md
@@ -39,9 +39,9 @@ const storage = createStorage({
 });
 ```
 
-Usage with Redis cluster (e.g. AWS ElastiCache or Azure Redis Cache):
+Usage with a Redis cluster (e.g. AWS ElastiCache or Azure Redis Cache):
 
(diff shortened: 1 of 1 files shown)


cjpearson:
One idea for the warning section:

> If you connect to a cluster, when running commands that operate over multiple keys, all keys must be part of the same hashslot. Otherwise you may encounter the Redis error `CROSSSLOT Keys in request don't hash to the same slot`. You should use [`hashtags`](https://redis.io/docs/latest/operate/oss_and_stack/reference/cluster-spec/#hash-tags) to control how keys are slotted. If you want all keys to hash to the same slot, you can include the hashtag in the base prefix by wrapping it in curly braces. Read more about [Clustering Best Practices](https://redis.io/blog/redis-clustering-best-practices-with-keys/).

pi0 (approved):
Thnx ❤️

PR #594, opened 2025-02-19
chore: de-dupe rollup

43081j:
There are currently multiple copies of rollup in our dependency tree. This de-dupes into a single version.


pi0:
a022bafa8e034893294f4ed6d7dd3d3f1cc36175 (FYI i use [codeup](https://github.com/unjs/codeup) for periodic cleanups)

PR #595, opened 2025-02-19
chore: fix corepack CI

43081j:
Node 20 ships with an old corepack that causes all sorts of nonsense right now

i saw pi0 has already been in the threads around it too

for now, we can setup pnpm directly and switch back once node 20.x ships with an upgraded corepack

cc @pi0


pi0:
91e2d7db6e922efe5256a77737c04bcbe9be76a6

(Github/node has not responded this late to this important issue I feel it is an intentional act to force the ecosystem out of using corepack..)

PR #596, opened 2025-02-21
refactor: upgrade to  chokidar to v4

pi0:
We initially upgraded to chokidar v4 (#489) then reverted back to #502 as a hotfix (see relevant issues)

Stability and ecosystem adoption seem now much better. The only remaining part is that chokidar v4 dropped glob support. 

This PR updates to use same `anymatch` matcher we use for listing key filtering, to also apply to `watchOptions.ignored`

- ~~If users already provided custom `ignore` option, it would be breaking change: Added via a lightweight polyfill with pathe~~
- ~~Default ignores `["**/node_modules/**", "**/.git/**"]` added using simply regex matcher~~

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,7 +51,6 @@
     "lru-cache": "^10.4.3",
     "node-fetch-native": "^1.6.6",
     "ofetch": "^1.4.1",
-    "pathe": "^2.0.3",
     "ufo": "^1.5.4"
   },
   "devDependencies": {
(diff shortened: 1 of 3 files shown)


43081j:
Looks good to me!

danielroe (approved):

pi0:
(updated impl to reuse same anymatch we use for filtering)

PR #600, opened 2025-03-12
feat(redis, vercel-kv, upstash): use non-blocking methods

### LINKED ISSUE #418: [redis] use `scan` to iterate keys
### Describe the feature

Working with large redis db, you might encounter an error if you try to fetch too many keys altogether.
There should be a scan option available (or getKeys should be implemented via scan)

### Additional information

dschewchenko:
resolves #418

These changes follow official Redis recommendations for production environments and improve handling of large datasets.

- Replaced blocking del with non-blocking unlink for better performance
- Implemented cursor-based scan instead of keys to prevent server blocking

### DIFF
--- a/src/drivers/redis.ts
+++ b/src/drivers/redis.ts
@@ -32,11 +32,16 @@ export interface RedisOptions extends _RedisOptions {
    * Default TTL for all items in seconds.
    */
   ttl?: number;
+
+  /**
+   * How many keys to scan at once.
+   *
+   * [redis documentation](https://redis.io/docs/latest/commands/scan/#the-count-option)
+   */
+  scanCount?: number;
 }
 
 const DRIVER_NAME = "redis";
-
-const batchSize = 100;
 
 export default defineDriver((opts: RedisOptions) => {
   let redisClient: Redis | Cluster;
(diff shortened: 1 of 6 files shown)


src/drivers/redis.ts
```diff
@@ -55,6 +57,22 @@ export default defineDriver((opts: RedisOptions) => {
...
   const d = (key: string) => (base ? key.replace(base, "") : key); // Deprefix a key
+  // Helper function to scan all keys matching a pattern
+  const scanAllKeys = async (pattern: string): Promise<string[]> => {
+    const client = getRedisClient();
+    const keys: string[] = [];
+    let cursor = "0";
+
+    do {
+      const [nextCursor, scanKeys] = await client.scan(
+        cursor, "MATCH", pattern, "COUNT", batchSize
```
pi0, line 68: The default count size is `10` (less blocking and mem footprint), wondering if we should leave it as as and instead introduce a configuration to customize

dschewchenko: hey! interesting question. I think to make this customizable in next major version

pi0: https://github.com/unjs/unstorage/pull/600/commits/39a8617f643d3c00116d391338b1932e9cfcbca9

pi0 (approved):
Thanks!

src/drivers/redis.ts
```diff
@@ -57,20 +62,19 @@ export default defineDriver((opts: RedisOptions) => {
...
     const keys: string[] = [];
     let cursor = "0";
-
     do {
-      const [nextCursor, scanKeys] = await client.scan(
-        cursor, "MATCH", pattern, "COUNT", batchSize
-      );
+      const [nextCursor, scanKeys] = opts.scanCount
+        ? await client.scan(cursor, "MATCH", pattern, "COUNT", opts.scanCount)
+        : await client.scan(cursor, "MATCH", pattern);
```
dschewchenko, line 74: How about simplify this to default redis offset option **10** 

> The default COUNT value is 10.

```
const [nextCursor, scanKeys] = await client.scan(cursor, "MATCH", pattern, "COUNT", opts.scanCount || 10);
```

It's shorter than ternary expression =)

pi0: From what i briefly understand from docs, there is no guarantee on count value. In case there would be different default, I think not hardcoding `10` might be best.

Issue #573, opened 2025-01-15
Cloudflare cache API 


cosbgn:
### Describe the feature

Is there a way to use the cloudflare cache api? I deploy nuxt with cloudflare_module preset (workers).
https://developers.cloudflare.com/workers/runtime-apis/cache

### Additional information

- [ ] Would you be willing to help implement this feature?

pi0:
CachedStorage is not suitable for KV storage.

Lets track with https://github.com/nitrojs/nitro/issues/2124

PR #603, opened 2025-03-19
feat: `cloudflare-cache-binding` driver

### LINKED ISSUE #602: Cloudflare Workers Cache API driver
### Describe the feature

A driver for interfacing with the Cloudflare Workers Cache runtime API

- https://developers.cloudflare.com/workers/runtime-apis/cache/
- https://developers.cloudflare.com/workers/examples/cache-api/

The API allows:
- `getItem`
- `hasItem` via `getItem`
- `setItem`
- `removeItem`

`getItems` is not supported using Cloudflare's implementation of the cache API.

### Additional information

RihanArfan:
resolves #602, #573

TODO:
- [ ] Needs testing
- [ ] Allow setting cache tag
    - https://developers.cloudflare.com/cache/how-to/purge-cache/purge-by-tags/
    - https://blog.cloudflare.com/instant-purge-for-all/


RihanArfan:
If Nitro v3 can do cached event handlers and functions directly via the cache API then I think may not be necessary then 😄 

Happy for you to close if Nitro is handling it better.

pi0:
@RihanArfan any update on this? (have you tested this solution on prod?)

ascorbic:
This isn't Cloudflare-specific. It's using the `CacheStorage` api, which is also [supported by other runtimes too](https://runtime-compat.unjs.io/#CacheStorage)

RihanArfan:
@ascorbic Cloudflare's implementation of the Cache API is slightly different from the web standard - https://developers.cloudflare.com/workers/runtime-apis/cache/#accessing-cache

Most notably, there's no way to list all keys - https://developer.mozilla.org/en-US/docs/Web/API/CacheStorage/keys

@pi0 Unfortunately haven't been able to yet sorry

ascorbic:
@RihanArfan is there a way to implement this in a cross-platform way though, that supports runtimes that implement the web standard too

eduardoboucas:
> @RihanArfan is there a way to implement this in a cross-platform way though, that supports runtimes that implement the web standard too

Netlify would also be interested in this.

dario-piotrowicz:
> Disabled tests with a manual test for now (works in prod sandbox with using `unstorage://` as cache key)
> 
> @dario-piotrowicz @petebacondarwin could you please kindly review this PR and confirm if i am not mistaking and wrangler dev proxy does not implement caches API?

I am not very familiar with this codebase unfortunately, but by quickly skimming through the PR it all seems correct to me 🙂 

Regarding the cache not being implemented you're right, we also have an issue for this: https://github.com/cloudflare/workers-sdk/issues/4879 (which refers to `getBidingsProxy` the old name for `getPlatformProxy`, I'll update the issue with the correct name) unfortunately this piece of work has never been prioritized (also it being not trivial doesn't help). The fact that we're trying to get more and more code to run natively in our runtime also makes it a bit improbable in my opinion that we can prioritize `getPlatformProxy`-only features (unless there's a significant demand/motivation for it) 🥲 

(PS: Running code inside `miniflare` should support the caching API... although I just manually checked that and that doesn't seem to work... I'll need to investigate a bit 🤔)

pi0 (approved):
Lets merge it for v2. Testing was manual but it seems a simple API we also should investigate netlify or other platform's support via a new generic driver i guess

adamdehaven:
Are there docs for the implementation?

montgomery1944:
@RihanArfan, @pi0, I've tested this implementation in Cloudflare Worker and it seems that it does not work, because the request URL used to store the cache entry is using custom scheme: "unstorage". During the testing the only two schemes that worked for me were: "http" and "https".

Please also check the note on https://developers.cloudflare.com/workers/runtime-apis/cache/ "When using the cache API, avoid overriding the hostname in cache requests, as this can lead to unnecessary DNS lookups and cache inefficiencies. Always use the hostname that matches the domain associated with your Worker.".

pi0:
IIRC last time testing in workers playground it was working. @montgomery1944 can you share the code you run in there?

montgomery1944:
@pi0, here is the code that I was testing:

```typescript
function cacheKey(key: string): string {
  return `https://my-worker.example/${key.replace(/:/g, "/")}`
}

async function readL1(key: string): Promise<string | null> {
  const response = await caches.default.match(cacheKey(key))
  return response ? response.text() : null
}

async function writeL1(key: string, value: string, ttl?: number): Promise<void> {
  const headers: Record<string, string> = {}
  if (ttl) {
    headers["Cache-Control"] = `max-age=${ttl}`
  }
  await caches.default.put(
      cacheKey(key),
      new Response(value, { headers })
  )
}

await writeL1("test", "test", 120);
const response = await readL1("test");
```

It works with "https" scheme, it does not work with "unstorage" scheme.

montgomery1944:
@pi0, I forgot to mention, that I wasn't testing this in the Workers Playground, but in a real worker deployed to Cloudflare.

medz:
Thank you for your hard work! I just discovered this and ran a real-world Workers test, only to find a bug that prevents it from being used: https://github.com/unjs/unstorage/issues/789

PR #610, opened 2025-04-07
feat: support `getItem(key, { type })`

jlucaso1:
Implements #528.

Adds an opt-in `type` option to `getItem()` / `getItems()` to read a value in a specific format, instead of always going through lenient JSON parsing.

```ts
const text = await storage.getItem("user:1", { type: "text" }); // string
const json = await storage.getItem("user:1", { type: "json" }); // JSONValue
const bytes = await storage.getItem("avatar.png", { type: "bytes" }); // Uint8Array
const blob = await storage.getItem("avatar.png", { type: "blob" }); // Blob
const stream = await storage.getItem("avatar.png", { type: "stream" }); // ReadableStream<Uint8Array>
```

| `type`   | Returns                      | Read from                 |
| -------- | ---------------------------- | ------------------------- |
| _(none)_ | unchanged (`destr()`)        | `driver.getItem`          |
| `text`   | `string`                     | `driver.getItem`          |
| `json`   | `JSONValue`                  | `driver.getItem`          |
| `bytes`  | `Uint8Array`                 | `driver.getItemRaw`       |
| `blob`   | `Blob`                       | `driver.getItemRaw`       |
| `stream` | `ReadableStream<Uint8Array>` | `driver.getItemRaw`       |

Missing keys return `null` for every type.

### Raw values

`bytes` / `blob` / `stream` read the driver's **raw** value and normalize whatever the driver emits — `Uint8Array`, `Buffer`, `ArrayBuffer`, `ArrayBufferView`, `string`, `Blob`, `Response`-like (`.arrayBuffer()`), `ReadableStream` and async iterables (Node.js streams). Drivers without `getItemRaw` fall back to `getItem` + `base64:` decoding.

Symmetrically, **any** `type` decodes raw values first, so `text` and `json` never surface the `base64:` wrapper (or a byte array) that drivers store raw values under:

```ts
await storage.setItemRaw("greeting", new TextEncoder().encode("hello"));

await storage.getItem("greeting", { type: "text" }); // "hello"
await storage.getItem("greeting", { type: "bytes" }); // Uint8Array
await storage.getItem("greeting"); // "base64:aGVsbG8=" (unchanged, no `type`)
```

`blob` and `stream` are pass-through when the driver already returns one — no buffering.

### `getItems`

`type` works per batch or per item, and the result value type is narrowed accordingly:

```ts
await storage.getItems(["avatar:1", "avatar:2"], { type: "bytes" });
// { key: string; value: Uint8Array | null }[]

await storage.getItems([
  { key: "user:1", options: { type: "json" } },
  { key: "avatar:1", options: { type: "bytes" } },
]);
// { key: string; value: JSONValue | Uint8Array | null }[]
```

`bytes` / `blob` / `stream` entries have no batch equivalent on the driver side, so they are resolved individually **in parallel with** the driver's `getItems()` call for the remaining keys, then merged back into input order.

### Types

- `getItem` gains an overload keyed on `type`, so `{ type: "bytes" }` returns `Promise<Uint8Array | null>`. Without `type`, the existing `StorageItemMap` / generic overloads are unchanged.
- `getItems` infers per-item types from a literal input array, falling back to the common `type` and then to the storage value type. Non-literal `string[]` inputs keep the common `type`.
- New exported types: `GetItemType`, `GetItemTypeMap`, `GetItemOptions`, `GetItemsInput`, `JSONValue`, `JSONObject`, `JSONArray`.
- Invalid types are rejected at compile time (`{ type: "buffer" }` errors).

### Notes

- Fully backwards compatible: with no `type`, `getItem` / `getItems` behave exactly as before, except that a driver returning `undefined` now consistently yields `null`.
- Conversion helpers (`toBytes` / `toBlob` / `toStream`) live in `src/_utils.ts`; the previously proposed custom binary transformer and the `undio` dependency were both dropped.
- Docs updated in `docs/1.guide/1.index.md`; tests cover every conversion, each raw input shape, mixed batches, mount ordering, error cases and reader-lock release, plus type tests.


src/storage.ts
```diff
@@ -473,7 +473,45 @@ export function createStorage<T extends StorageValue>(
     },
     // Aliases
     keys: (base, opts = {}) => storage.getKeys(base, opts),
-    get: (key: string, opts = {}) => storage.getItem(key, opts),
+    get: async (
```
pi0, line 476: get is an alias to getItem, they should behave identical

src/storage.ts
```diff
@@ -172,12 +172,44 @@ export function createStorage<T extends StorageValue>(
...
       const { relativeKey, driver } = getMount(key);
-      return asyncCall(driver.getItem, relativeKey, opts).then(
-        (value) => destr(value) as StorageValue
-      );
+
+      const type = opts.type;
+
+      if (type === "bytes" || type === "stream" || type === "blob") {
+        if (driver.getItemRaw) {
+          return asyncCall(driver.getItemRaw, relativeKey, opts);
```
pi0, line 188: Here we should be careful and do data transforms, raw value is not guaranteed to be same as type.

pi0: same for L191

jlucaso1: I added an utility function to transform this values

src/utils.ts
```diff
@@ -124,3 +124,85 @@ export function filterKeyByBase(
...
+ */
+export function transformRawToType(
+  raw: any,
+  type: "bytes" | "blob" | "stream"
+) {
+  // Handle "bytes"
+  if (type === "bytes") {
+    if (
+      raw instanceof Uint8Array ||
+      (typeof Buffer !== "undefined" && raw instanceof Buffer)
```
pi0, line 139: Node buffer extends `Uint8Array` this check can be removed

pi0:
After pushed some simplification refactors (5b020ecbfbd1010b62710553ae67bffabe720d11) i noticed there are cases we don't cover (for example blob to bytes) and cases we cover but without best perf (for example blob can be constructed from string and also directly converted to an stream).

Thinking to migrate to [undio](https://github.com/unjs/undio)

jlucaso1:
@pi0 The last commit is an attempt of migrate to undio. Can you review again?

pi0 (approved):
finally! thanks for pr ❤️

PR #611, opened 2025-04-07
fix(prefixStorage): prefix `getItems` and `setItems`

### LINKED ISSUE #396: `getItems` does not work with `prefixStorage`

jlucaso1:
Enhance the `prefixStorage` function to correctly handle `getItems` and `setItems` by applying the prefix to keys. This resolves the issue where `getItems` was returning null for prefixed keys. Additional tests ensure that both `getItems` and `setItems` function as expected with prefixed keys.

Fixes #396


src/utils.ts
```diff
@@ -44,6 +44,44 @@ export function prefixStorage<T extends StorageValue>(
       // Remove Prefix
       .then((keys) => keys.map((key) => key.slice(base.length)));
 
+  if (typeof storage.getItems === "function") {
+    nsStorage.getItems = <U extends T>(
+      items: (string | { key: string; options?: any })[],
+      commonOptions?: any
```
43081j, line 50: both `options` and `commonOptions` are `TransactionOptions`. so you may as well use the real type here instead of `any`

src/utils.ts
```diff
@@ -44,6 +44,44 @@ export function prefixStorage<T extends StorageValue>(
...
+          value: entry.value,
+        }))
+      );
+    };
+  }
+
+  if (typeof storage.setItems === "function") {
+    nsStorage.setItems = <U extends T>(
+      items: { key: string; value: U; options?: any }[],
+      commonOptions?: any
```
43081j, line 74: again these too can be `TransactionOptions`

test/storage.test.ts
```diff
@@ -249,4 +249,43 @@ describe("Regression", () => {
       await storage.clear();
     }
   });
+
+  it("prefixStorage getItems returns null (issue #396)", async () => {
```
43081j, line 253: is this test description wrong/old? 👀 

since it seems we're testing that `getItems` returns values and non-prefixed keys

43081j (commented):
looks good to me 👍 

will leave pooya to approve/merge

test/storage.test.ts
jlucaso1: It's old. I used this test to replicate the bug

pi0 (approved):
Thanks ❤️

PR #612, opened 2025-04-07
Prevent mutation of input options in all drivers

### LINKED ISSUE #566: Make sure drivers don't upsdate input options object
### Describe the feature

Found from https://github.com/nitrojs/nitro/issues/2174

Drivers such as fs/fsLite currently modify options objects. this might have unwanted side-effects especially if input options are used in different places.

We should:
- Make `defineDriver` types to make sure options object is (deep) ReadOnly
- Possibly update tests to make sure there is no runtime update
- Fix drivers that currently modify input to instead, make a copy of object

### Additional information

jlucaso1:
Ensure that input options are not mutated across all drivers by creating a new options object from the input. This change enhances the integrity of the options passed to each driver.
Resolves #566

### DIFF
--- a/src/drivers/azure-app-configuration.ts
+++ b/src/drivers/azure-app-configuration.ts
@@ -36,7 +36,8 @@ export interface AzureAppConfigurationOptions {
 
 const DRIVER_NAME = "azure-app-configuration";
 
-export default defineDriver((opts: AzureAppConfigurationOptions = {}) => {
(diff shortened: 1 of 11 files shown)


43081j:
can you show us where we were mutating the options?

I wonder if we should treat them as read-only and then only clone them in places where we _must_ mutate

since I suspect most drivers don't mutate, so can avoid having to allocate a new object and could let the TS compiler catch disallowed mutations (it'd be read-only deeply)

PR #614, opened 2025-04-10
feat(redis): add support for `preConnect`

### LINKED ISSUE #469: [redis] support `preConnect` option
### Environment

Unstorage v1.10.2

### Reproduction

1. Create a Driver or Storage object without manually calling getRedisClient().
2. Attempt to call setItem or getItem.

### Describe the bug

The Redis instance is initialized only when the `getRedisClient()` function is called. However, `getRedisClient()` is never invoked during the creation of the `Driver` or the `Storage` objects. As a result, the first call to `getRedisClient()` occurs only when `setItem` or `getItem` is called.

This causes an issue where the Redis instance is created and attempts to get or set an item, but the `ioredis` instance has not yet established a connection to the Redis server.

### Expected Behavior

The Redis instance should be initialized and connected to the Redis server before any get or set operations are performed.

### Actual Behavior

The Redis instance is created at the time of the first `setItem` or `getItem` call, and the `ioredis` instance is not yet connected.

### Possible Solution

Ensure that the `getRedisClient()` function is called after beeing defined to establish a connection to the Redis server before any operations are performed.
From what I can see, it seems to be same for Mongodb

### Additional context

_No response_

### Logs

_No response_

kricsleo:
resolves #469

This PR adds `preConnect` support for the Redis driver.

It also disables `lazyConnect` by default (if unset) when `preConnect` is enabled for pre-connection.

A roughly benchmark indicates that `preConnect` slightly speeds up the initial read/write operations. (However, it may also increase the application startup time.)

image


kricsleo:
> ioredis lazyConnect is false by default do we need to do it again? (seems logic can be simplified)
Yes, I confirmed it [defaults to `false`](https://github.com/redis/ioredis/blob/2ed6414339bdda3e427ed1e8484263cc4848f2ac/lib/redis/RedisOptions.ts#L238). Will remove the extra code in this PR.

Then this might be misleading:
https://github.com/unjs/unstorage/blob/main/docs/2.drivers/redis.md?plain=1#L76

docs/2.drivers/redis.md
```diff
@@ -70,10 +70,11 @@ const storage = createStorage({
 - `cluster`: List of redis nodes to use for cluster mode. Takes precedence over `url` and `host` options.
 - `clusterOptions`: Options to use for cluster mode.
 - `ttl`: Default TTL for all items in **seconds**.
+- `preConnect`: Whether to initialize the redis instance immediately. Otherwise, it will be initialized on the first read/write call. When enabled, it also sets `lazyConnect` to `false` if unset. Default disabled.
 
 See [ioredis](https://github.com/luin/ioredis/blob/master/API.md#new-redisport-host-options) for all available options.
 
-`lazyConnect` option is enabled by default so that connection happens on first redis operation.
+However, if `preConnect` is enabled while `lazyConnect` is unset, lazy connection will be disabled in favor of pre-connection.
```
pi0, line 77:
```suggestion
If `preConnect` is enabled, driver tries to connect as soon as initialized.
```

pi0 (approved):
Thanks!

PR #616, opened 2025-04-11
fix!: fallback to the root mountpoint when no matching mountpoints found

### LINKED ISSUE #336: Inconsistency between clear and getKeys
### Environment

unstorage@1.9.0, node v20.9.0

### Reproduction

Currently `getKeys` and `clear` behave differently when they are passed a prefix.

```
const prefix = 'test';
let keys = await storage.getKeys(prefix);
console.log(keys); // ['test:123', 'test:abc']
await storage.clear(prefix);
keys = await storage.getKeys(prefix);
console.log(keys); // ['test:123', 'test:abc']
```

`clear` seems to ignore the prefix if it is not a mount, while `getKeys` finds all keys which begin with the prefix. Since the options are named the same I had assumed they would behave similarly.

### Describe the bug

Is this behavior intentional? Currently I'm working around it by individually deleting keys, but it would be nice if `clear` also worked.

```
const prefix = 'test';
let keys = await storage.getKeys(prefix);
console.log(keys); // ['test:123', 'test:abc']
await Promise.all(keys.map((k) => this.storage.removeItem(k)));
keys = await storage.getKeys(prefix);
console.log(keys); // []
``` 

### Additional context

_No response_

### Logs

_No response_

kricsleo:
resolves #336

### DIFF
--- a/src/storage.ts
+++ b/src/storage.ts
@@ -59,20 +59,31 @@ export function createStorage<T extends StorageValue>(
   };
 
   const getMounts = (base: string, includeParent?: boolean) => {
-    return context.mountpoints
-      .filter(
-        (mountpoint) =>
-          mountpoint.startsWith(base) ||
-          (includeParent && base!.startsWith(mountpoint))
-      )
-      .map((mountpoint) => ({
-        relativeBase:
-          base.length > mountpoint.length
-            ? base!.slice(mountpoint.length)
-            : undefined,
-        mountpoint,
-        driver: context.mounts[mountpoint]!,
-      }));
+    const mountpoints = context.mountpoints.filter(
+      (mountpoint) =>
+        mountpoint.startsWith(base) ||
+        (includeParent && base!.startsWith(mountpoint))
+    );
+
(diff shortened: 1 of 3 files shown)


pi0 (approved):
Thanks, this is a valid fix however since risky (some usage might have been depending on faulty behavior and this can cause data loss) we should do it as a major version.

ToBinio:
Hi,
I just did a small deep dive into the clear function because of https://github.com/unjs/unstorage/pull/642#issue-3203056678 and https://github.com/unjs/unstorage/issues/638#issuecomment-3038546169

Is it really the best idea to delete the whole DB if I enter a too precise prefix? Isn't that kinda backwards?
I also don't think this solves #336 since the behaviors still differ... `getKeys` returns the expected subset and `clear` nukes the whole storage

---

https://github.com/unjs/unstorage/blob/25192508fd81e204ed55086cec583ca890973d7c/src/storage.ts#L387

I would suggest having `getMounts(base, true)` in clear and letting the drivers handle only deleting a subset of data.
Some drivers e.g. redis would already handle that

PR #622, opened 2025-04-15
fix(fs): exclude ignored file in key listing

### LINKED ISSUE #621: Ignored files are not excluded from key listing
### Environment

Tested with unstorage v1.15.0

### Reproduction

```js
import { createStorage } from "unstorage";
import fsDriver from "unstorage/drivers/fs";
import { writeFile, mkdir } from "fs/promises";
import { join } from "path";

const testFolders = join("dir1", "dir2");
await mkdir(testFolders, { recursive: true });

const storage = await createStorage({
  driver: fsDriver({
    base: "./dir1",
    ignore: "**/dir2/test.txt",
  }),
});

await writeFile("./dir1/dir2/test.txt", "");

console.log(await storage.getKeys());
```

This code outputs `[ 'dir2:test.txt' ]` but I would expect it to output `[]`.

### Describe the bug

When creating a `fsDriver` a `ignore`  patterns can be specified.

When listing keys with `getKeys`  files that match an ignore pattern are still returned in some cases.
This happens if the parent path of the file does not match an ignore pattern.

The ignore pattern still correctly excludes files from being watched for changes.
So it appears that the ignore patterns are applied correctly when setting up file watching, but not when applied for listing keys.

### Additional context

_No response_

### Logs

```sh

```

odzhychko:
The ignore pattern was applied only to the file name and not to the full file path.
This caused files that should have been ignored to appear in the key listing.

Resolves #621

### DIFF
--- a/src/drivers/utils/node-fs.ts
+++ b/src/drivers/utils/node-fs.ts
@@ -69,7 +69,7 @@ export async function readdirRecursive(
           files.push(...dirFiles.map((f) => entry.name + "/" + f));
         }
       } else {
-        if (!(ignore && ignore(entry.name))) {
+        if (!(ignore && ignore(entryPath))) {
           files.push(entry.name);
         }
       }
--- a/test/drivers/fs.test.ts
+++ b/test/drivers/fs.test.ts
@@ -1,8 +1,9 @@
-import { describe, it, expect, vi } from "vitest";
+import { describe, it, expect, vi, afterEach } from "vitest";
 import { resolve } from "node:path";
 import { readFile, writeFile } from "../../src/drivers/utils/node-fs";
-import { testDriver } from "./utils";
+import { testDriver, type TestContext } from "./utils";
 import driver from "../../src/drivers/fs";
+import { createStorage } from "../../src";
 
 describe("drivers: fs", () => {
   const dir = resolve(__dirname, "tmp/fs");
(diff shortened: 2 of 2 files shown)


src/drivers/utils/node-fs.ts
```diff
@@ -69,7 +69,7 @@ export async function readdirRecursive(
           files.push(...dirFiles.map((f) => entry.name + "/" + f));
         }
       } else {
-        if (!(ignore && ignore(entry.name))) {
+        if (!(ignore && ignore(entryPath))) {
```
pi0, line 72: I have a feeling some old patterns could only cover `entry.name` but not sure without testing. to be on safe side, we could check both.

odzhychko: The fix as proposed is indeed breaking for some patterns. 

For example, the pattern `"ignored.txt"` currently would exclude the file `"{base}/dir/ignored.txt"` from key listings. But after the fix, the pattern would need to be changed to `"**/ignored.txt"`.

One could argue that excluding `"{base}/dir/ignored.txt"` because of the pattern `"ignored.txt"`  is also a bug (with the same cause as #621) because this pattern does not exclude the file from being watched. The following example shows the pattern having an effect on `storage.getKeys` but not on `storage.watch`.

```js
import { createStorage } from "unstorage";
import fsDriver fSo I think, "entry.name" could be checked for backward compatibility. 
But I guess that is something for you as a maintainer to decide. rom "unstorage/drivers/fs";
import { writeFile, mkdir } from "fs/promises";
import { join } from "path";

const testFolders = join("base", "dir");
await mkdir(testFolders, { recursive: true });

const storage = await createStorage({
  driver: fsDriver({
    base: "./base",
    ignore: [
      "ignored.txt"
    ]
  }),
});

const unwatch = await storage.watch((event, key) => {
  console.log(event, key)
});

await writeFile("./base/dir/ignored.txt", "test");
await writeFile("./base/dir/notIgnored.txt", "test");
// "update dir:notIgnored.txt" is logged.
// "update dir:ignored.txt" is not logged.


console.log(await storage.getKeys())
// "[ 'dir:notIgnored.txt' ]"
await unwatch()
```

Whether what I describe can be considered a bug or just an inconsistent feature worth maintaining backwards compatibility for, is something for you to decide. Please let me know what you decided, and I will make the appropriate change.

My option is that backwards compatibility could be maintained without impeding intended usage. Usually, patterns should always start with `"**"` or `"/"` because the ignore patterns are applied to an absolute path.  But if someone nevertheless just used the file name (e.g., `"ignored.txt"`) and it worked, the current behavior could be maintained (maybe until the next major version) by checking `entry.name` in addition to `entryPath`.

pi0: lets have it as breaking change

PR #623, opened 2025-04-22
fix(session-storage): correct `name`

juliomrqz:
I'm working on a custom implementation that falls back to a list of drivers when the main driver are not supported. For example, localStorage might be unavailable or unreliable in some mobile devices, so sessionStorage is a second option.

I need to log the name of the driver being supported, but I noticed that  the name of the `session-storage` is replaced by `localstorage` name:

CleanShot 2025-04-22 at 12 43 09@2x

This PR defines correctly the driver name of session-storage.


43081j:
good catch!

definitely a bug. i wonder if we do this anywhere else too 🤔

pi0 (approved):
thanks!

PR #625, opened 2025-04-23
feat(mongodb): allow setting additional clientOptions

alexbevi:
The PR addresses #626 by incorporating MongoDB's [wrapping client library](https://github.com/mongodb/specifications/blob/master/source/mongodb-handshake/handshake.md#supporting-wrapping-libraries) specification for the connection handshake to allow library details to be included in the metadata written to `mongos` or `mongod` logs.

For example, this change would allow server-side logs such as the following:

> {"t":{"$date":"2024-01-27T23:10:40.108Z"},"s":"I","c":"NETWORK","id":51800,"ctx":"conn16235","msg":"client metadata","attr":{"remote":"127.0.0.1:1094","client":"conn16235","doc":{"driver":{`"name":"nodejs|unstorage"`,`"version":"4.16.0"`},"platform":"Node.js v18.18.2, LE","os":{"name":"linux","architecture":"x64","version":"5.15.133+","type":"Linux"}}}}

For anyone hosting clusters with connections coming from different applications this can help differentiate connections and facilitate log analysis.


src/drivers/mongodb.ts
```diff
@@ -29,7 +29,11 @@ export default defineDriver((opts: MongoDbOptions) => {
       if (!opts.connectionString) {
         throw createRequiredError(DRIVER_NAME, "connectionString");
       }
-      const mongoClient = new MongoClient(opts.connectionString);
+      const mongoOptions: any = {};
+      mongoOptions.driverInfo = {
+        name: "unstorage",
+      };
+      const mongoClient = new MongoClient(opts.connectionString, mongoOptions);
```
pi0, line 36: I think we should instead support generic `mongoOptions` as a driver option so you can set it to `{ driverInfo: { name: "..." } }` based on your needs

43081j: i agree. we could default if it helps:

```ts
// in the top of the file
const defaultMongoOptions: MongoClientOptions = {
  driverInfo: {
    name: "unstorage"
  }
};

// here (where i'm commenting in this PR)
const mongoOptions: MongoClientOptions = {
  ...defaultMongoOptions, // don't bother deep-merging, people can override `driverInfo` as a whole
  ...opts.clientOptions
};
```

then update the `MongoDbOptions` to have:

```ts
clientOptions?: MongoClientOptions;
```

43081j:
Looks good to me 👍

@pi0 I'll leave you to merge once you've had time to read through

pi0 (approved):
Thanks for the PR. I have updated to only allow customizing options to set any desired name for the logs visibility.

Issue #629, opened 2025-05-05
Typescript error with prefixStorage generics


ondrejvelisek:
### Environment

I dont think any of those matters, but just anyway Im adding my env deps (EDIT: it does, see EDIT about unstorage versions below):

node --version                                                                                                                                                                               
v20.13.1

"packageManager": "pnpm@9.5.0

"unstorage": "^1.16.0",

"typescript": "^5.8.3",

"@upstash/redis": "^1.34.8",

"@vercel/kv": "^3.0.0",

### Reproduction

See this simple sandbox. I just copied docs example:
https://codesandbox.io/p/sandbox/jwtrzs

### Describe the bug

In a simple example with prefixStorage I get this TS error:

Argument of type 'Storage<StorageValue>' is not assignable to parameter of type 'Storage<string>'.
  The types returned by 'getItem(...)' are incompatible between these types.
    Type 'Promise<StorageValue>' is not assignable to type 'Promise<string | null>'.
      Type 'StorageValue' is not assignable to type 'string | null'.
        Type 'number' is not assignable to type 'string'. typescript(2345)

I dont want to use casting of storage (as keyword) since it is not typesafe. Nor want to type the storage itself because I want to be able to use it with different (incompatible) prefixes.

### Logs

```sh

```

EDIT: 

I tried several versions of unstorage.

versions working correctly:
[1.13.1](https://www.npmjs.com/package/unstorage/v/1.13.1)
[1.14.0](https://www.npmjs.com/package/unstorage/v/1.14.0)

versions with the described type error:
[1.14.1](https://www.npmjs.com/package/unstorage/v/1.14.1)
...
[1.14.4](https://www.npmjs.com/package/unstorage/v/1.14.4)
[1.15.0](https://www.npmjs.com/package/unstorage/v/1.15.0)
[1.16.0](https://www.npmjs.com/package/unstorage/v/1.16.0)

So the problem is with the changes introduced between v1.14.0 and v1.14.1
It should help with finding the issue.

EDIT2: 
PR Bugfix 532 probably introduced the bug:
https://github.com/unjs/unstorage/pull/532

Aslemammad:
we started getting this too on https://github.com/stackblitz-labs/pkg.pr.new/blob/main/packages/app/server/utils/bucket.ts

apparently, it was working in the previous versions but right now it's raising type errors.

TheDutchCoder:
[...]

PR #526, opened 2024-12-18
build: generate drivers meta

pi0:
Auto-generate meta data for `builtinDrivers`, `BuiltinDriverName` and `BuiltinDriverOptions` exports.

This PR adds consistent kebab-case + camelCase variants to all exports.

Another change by this PR is that we don't double bundle driver options types anymore and types are reused from `unstorage/drivers/` making dist size almost half!

<details>

```ts

// Auto-generated using scripts/gen-drivers.
// Do not manually edit!
import type { AzureAppConfigurationOptions as AzureAppConfigurationOptions } from "unstorage/drivers/azure-app-configuration";
import type { AzureCosmosOptions as AzureCosmosOptions } from "unstorage/drivers/azure-cosmos";
import type { AzureKeyVaultOptions as AzureKeyVaultOptions } from "unstorage/drivers/azure-key-vault";
import type { AzureStorageBlobOptions as AzureStorageBlobOptions } from "unstorage/drivers/azure-storage-blob";
import type { AzureStorageTableOptions as AzureStorageTableOptions } from "unstorage/drivers/azure-storage-table";
import type { CapacitorPreferencesOptions as CapacitorPreferencesOptions } from "unstorage/drivers/capacitor-preferences";
import type { KVOptions as CloudflareKVBindingOptions } from "unstorage/drivers/cloudflare-kv-binding";
import type { KVHTTPOptions as CloudflareKVHttpOptions } from "unstorage/drivers/cloudflare-kv-http";
import type { CloudflareR2Options as CloudflareR2BindingOptions } from "unstorage/drivers/cloudflare-r2-binding";
import type { DB0DriverOptions as Db0Options } from "unstorage/drivers/db0";
import type { DenoKvNodeOptions as DenoKVNodeOptions } from "unstorage/drivers/deno-kv-node";
import type { DenoKvOptions as DenoKVOptions } from "unstorage/drivers/deno-kv";
import type { FSStorageOptions as FsLiteOptions } from "unstorage/drivers/fs-lite";
import type { FSStorageOptions as FsOptions } from "unstorage/drivers/fs";
import type { GithubOptions as GithubOptions } from "unstorage/drivers/github";
import type { HTTPOptions as HttpOptions } from "unstorage/drivers/http";
import type { IDBKeyvalOptions as IndexedbOptions } from "unstorage/drivers/indexedb";
import type { LocalStorageOptions as LocalstorageOptions } from "unstorage/drivers/localstorage";
[...]

PR #631, opened 2025-05-22
build: fix invalid runtime type

danielroe:
this resolves the issue reported [in this comment](https://github.com/unjs/unstorage/issues/629#issuecomment-2898846132) which prevented types from working in unstorage v1.16 at all

I think possibly https://github.com/unjs/unstorage/pull/526 _exposed_ it

~~to avoid recurrence, I've included an additional test step (and ignored the fact for the moment that in node16 bundle mode, the exports are [masquerading as cjs](https://github.com/arethetypeswrong/arethetypeswrong.github.io/blob/main/docs/problems/FalseCJS.md#-masquerading-as-cjs) - that needs to be resolved too but touches a slightly bigger surface area than this PR)~~

### DIFF
--- a/.attw.json
+++ b/.attw.json
@@ -1,5 +1,3 @@
 {
-  "ignoreRules": [
-    "false-cjs"
-  ]
+  "ignoreRules": ["false-cjs"]
 }
--- a/.attw.json
+++ b/.attw.json
@@ -1,0 +1,5 @@
+{
+  "ignoreRules": [
+    "false-cjs"
+  ]
+}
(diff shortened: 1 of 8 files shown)


pi0:
Thanks for the PR, but please revert ATW usage.

We do not encourage CJS, and node10 style will be dropped in all new packages. I don't want it in workflow unless it can be only enabled for node16/ESM (or we can build one for unbuild/obuild integration)

pi0 (approved):

PR #640, opened 2025-06-25
refactor: allow `@netlify/blobs` v9 and v10 as peer dependency

serhalp:
Breaking changes are compatible with unstorage:
- v9: https://github.com/netlify/primitives/releases/tag/blobs-v9.0.0
- v10: https://github.com/netlify/primitives/releases/tag/blobs-v10.0.0

This also bumps the dev dep to latest while I'm at it.


pi0:
Pushed commit to fix types

serhalp:
Yeah, I guess I need to adjust the netlify blobs driver options type to support some new options added in v10. I'll get to that soon.

src/drivers/netlify-blobs.ts
```diff
@@ -88,11 +89,20 @@ export default defineDriver((options: NetlifyStoreOptions) => {
       // @ts-expect-error has trouble with the overloaded types
       return getClient().get(key, { type: topts?.type ?? "arrayBuffer" });
     },
-    setItem(key, value, topts?: SetOptions) {
-      return getClient().set(key, value, topts);
+    async setItem(key, value, topts?: SetOptions) {
+      // NOTE: this returns either Promise<void> (pre-v10) or Promise<WriteResult> (v10+)
+      // TODO(serhalp): Allow drivers to return a value from `setItem`. The @netlify/blobs v10
+      // functionality isn't usable without this.
```
serhalp, line 95: Let's get this merged and released as is for now, but it would be great to discuss this afterwards!

serhalp:
@pi0 I saw you changed this to `build:`. shouldn't this be `fix`?

src/drivers/netlify-blobs.ts
serhalp: Everything will "work" fine—it just won't be possible to leverage [the new functionality in v10](https://github.com/netlify/primitives/pull/303) because the user won't be able to access the return value `{modified, etag}` from `set()` and `setJSON()`.

I think this is just a driver interface typing issue. It's marked as returning `Promise<void>` but we would need it to be something like `<T = void>...Promise<T>`.

PR #641, opened 2025-06-25
perf(mongoDB): use bulk operations for `getItems` and `setItems`

ToBinio:
This adds native implementations for the bulk operations `getItems` and `setItems` for the MongoDB driver.

---

This also re-enables the MongoDB tests - I had to use the beta version of `mongodb-memory-server`  [see](https://github.com/typegoose/mongodb-memory-server/issues/928) locally, so unsure how the CI reacts.

I also expanded the `getItems` test so it catches more edge cases
* unknown key → return null
* always return same order as keys

Please let me know if you have any problem with these changes and I will revert them!


src/drivers/mongodb.ts
```diff
@@ -56,17 +56,43 @@ export default defineDriver((opts: MongoDbOptions) => {
       const document = await getMongoCollection().findOne({ key });
       return document?.value ?? null;
     },
+    async getItems(items) {
+      const keys = items.map((item) => item.key);
+
+      const result = await getMongoCollection()
+        .find({ key: { $in: keys } })
```
pi0, line 63: Would it make sense if we use aggregation `$match` => `$getFields`? (might be faster if we let the server side keep result sorted)

pi0: Agreegate operations preserve order unlike find + `$in`.

I don't know why wrote getField yes project 😆

ToBinio: But would it not then preserve the order of how every the server stored it, so order of insertion and not order of user keys?

So this:

```js
  const result = await getMongoCollection()
    .aggregate([
      { $match: { key: { $in: keys } } },
      { $project: { key: 1, value: 1 } },
    ])
    .toArray();
```

Would not work.

Or am I still missing something?

pi0 (approved):

PR #642, opened 2025-07-04
perf(redis, upstash): use bulk operations for `getItems`

ToBinio:
This PR adds a native implementation for `getItems` for upstash and redis
`setItems` was not implemented since `mset` does not support setting a TTL, and I am unsure how to go about it.

sry for the random .then -> async changes let me know if I should revert them.

---

edit: removed incorrect note


pi0 (approved):
Thanks!

(as our upstash token is revoked i ran tests locally)

Issue #643, opened 2025-07-07
GitHub driver fails with 403 due to missing User-Agent header when used in Cloudflare Workers


alexandre-hallaine:
### Environment

When using the `unstorage` GitHub driver in a Cloudflare Worker, all fetch requests to `https://api.github.com` return a **403 Forbidden** error.

After debugging, the cause appears to be the **lack of a `User-Agent` header**, which GitHub's API requires. This issue only occurs when the driver runs in Cloudflare Workers or other serverless environments that do not automatically inject this header.

### Reproduction

Use the GitHub driver in a Worker or edge function:

```ts
await fetch("https://api.github.com/repos/alexandre-hallaine/ghstats/git/trees/main?recursive=1")
```

### Describe the bug

403 Forbidden
"Request forbidden by administrative rules. Please make sure your request has a User-Agent header."

### Additional context

```ts
$fetch(url, {
  headers: {
    "User-Agent": "unstorage-github-driver",
    ...options?.headers
  }
})
```

### Logs

```sh

```

PR #644, opened 2025-07-08
fix(driver/github): add User-Agent header for Cloudflare Workers to avoid 403

alexandre-hallaine:
This fixes a 403 Forbidden error when using the GitHub driver in Cloudflare Workers or similar serverless environments. GitHub's API requires a `User-Agent` header, which is missing in these environments by default.

This resolves issue #643.

### DIFF
--- a/src/drivers/github.ts
+++ b/src/drivers/github.ts
@@ -137,11 +137,10 @@ async function fetchFiles(opts: GithubOptions) {
       `/repos/${opts.repo}/git/trees/${opts.branch}?recursive=1`,
       {
         baseURL: opts.apiURL,
-        headers: opts.token
-          ? {
-              Authorization: `token ${opts.token}`,
-            }
-          : undefined,
+        headers: {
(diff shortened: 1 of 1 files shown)


alexandre-hallaine:
This code is tested

43081j:
makes sense 👍 

we can't really write a test for this without mocking out the underlying fetch call, so maybe will just have to do with manual testing for now

src/drivers/github.ts
```diff
@@ -137,11 +137,10 @@ async function fetchFiles(opts: GithubOptions) {
       `/repos/${opts.repo}/git/trees/${opts.branch}?recursive=1`,
       {
         baseURL: opts.apiURL,
-        headers: opts.token
-          ? {
-              Authorization: `token ${opts.token}`,
-            }
-          : undefined,
+        headers: {
+          "User-Agent": "unstorage-github-driver",
```
pi0, line 141:
```suggestion
          "User-Agent": "unstorage",
```

PR #647, opened 2025-07-12
chore: add monthly compatibility workflow

43081j:
Basically runs CI against various peers to ensure we're still
compatible.

@pi0 let me know if this makes sense. if you know a better way, i'm all ears. this is what i was talking about in #640.

### DIFF
--- a/.github/workflows/compat-tests.yml
+++ b/.github/workflows/compat-tests.yml
@@ -1,0 +1,41 @@
+name: Test Compatibility
+
+on:
+  schedule:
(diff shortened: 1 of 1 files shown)


pi0:
Intreating approach. What do you think if we instead add a script to check latest version of peer deeps (test next major in CI), it is a little hard to keep testing all older versions.

43081j:
similar to this but using `latest` of every peer?

we could probably assume older ones work since we must've tested them in the past. so maybe it is more useful testing against the latest ones

PR #652, opened 2025-08-11
docs(database): correct `tableName` config key

Kanashimo:
I've changed variable `table` to `tableName` because the first one don't exist.


pi0 (approved):
Thnx!

PR #653, opened 2025-08-12
docs(database): correct tableName config key (2)

Kanashimo:
Oopsie, I forgot to change `table` to `tableName` in options, sorry. (#652)


pi0 (approved):
Thanks!

PR #661, opened 2025-08-22
feat(drivers): add type generics to support strongly typed options of drivers

### LINKED ISSUE #660: Support strongly typed options in storage methods
### Describe the feature

It would be great if we could strongly type the options each driver expects in their methods which the storage passes onto the driver!

### Additional information

schplitt:
Implements support for typed options from driver. Resolves #660 

So I think I better explain what is going on here.

I added an optional type argument for the drivers to declare their option types (get, set, remove, has, getKeys).
This type argument is passed to the Storage (via the driver) where the actual type is then extracted and passed onto the methods.
It achives strongly typed options in the methods, for example `get` from the specific driver used.

I tried to deduce what kind of options are needed in what method, but please double check them in the `Driver` and `Storage` types.

As a side note, there is no type safety for the options inside the `createStorage` function, though as we currently always (as far as I tell) just pass them right back into the driver, there should be no issues.

As part of this I had to add these types to the `netlify-blob` driver. There you can see the strongly typed options for the `set`, `get` and `getKeys` in action.

From here we can go into each driver and set their option types to get strong type support in their option methods.

I hope my explanation makes sense.
Please take a look.
If you have any questions, please feel free to ask right away!

### DIFF
--- a/src/drivers/deno-kv.ts
+++ b/src/drivers/deno-kv.ts
@@ -12,6 +12,7 @@ export interface DenoKvOptions {
    */
   ttl?: number;
 }
+
 interface DenoKVSetOptions {
   /**
    * TTL in seconds.
@@ -75,12 +76,12 @@ export default defineDriver<DenoKvOptions, Promise<Deno.Kv | Kv>>(
         const value = await kv.get(r(key));
         return value.value;
       },
-      async setItem(key, value, tOptions: DenoKVSetOptions) {
+      async setItem(key, value, tOptions?: DenoKVSetOptions) {
         const ttl = normalizeTTL(tOptions?.ttl ?? opts?.ttl);
         const kv = await getKv();
         await kv.set(r(key), value, { expireIn: ttl });
       },
-      async setItemRaw(key, value, tOptions: DenoKVSetOptions) {
+      async setItemRaw(key, value, tOptions?: DenoKVSetOptions) {
         const ttl = normalizeTTL(tOptions?.ttl ?? opts?.ttl);
         const kv = await getKv();
         await kv.set(r(key), value, { expireIn: ttl });
--- a/src/drivers/netlify-blobs.ts
+++ b/src/drivers/netlify-blobs.ts
@@ -45,15 +45,20 @@ export interface NetlifyNamedStoreOptions
   deployScoped?: false;
 }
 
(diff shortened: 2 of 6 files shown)


pi0:
Thanks for the PR and explanations.

The main benefit of unstorage is that it abstracts away the implementation of drivers with a unified (same) interface. Adding per-driver extensions takes that benefit away, and is no better than directly using each driver's underlying client (it might be even more powerful and better typed!) 

Other than options for creating a driver, their low-level types should not vary.

I would like to instead shift towards introducing vendor namespaced options like `{ netlify: { ... }` + generic ones like `{ tll }`. 

We could try adding these extensions via a global shared interface in a non breaking change way.

schplitt:
Could you elaborate a bit on what you had in mind with the Picking and Omitting?

For me it sounds similar to what I came up with,but using the e.g. supported flags + driver name to later on pick/omit out of the global type.

To have type safety in the Storage methods this way we would still need to pass along the driver type into the storage to allow for type inference.

An example of your idea would be very helpful!

Anyway, I would like to help with that 😄

pi0:
Roughly what I have in mind: 

```ts
// [per driver]

export interface XYZOptions {}

export interface XYZDriver {
  setOptions: {};
  getOptions: {};
}

// [_drivers.ts (auto gen)]

export type BuiltinDrivers = {
  xyz: XYZDriver;
};

export type GetOptions = {
  xyz?: XYZDriver["getOptions"];
};

export type SetOptions = {
  xyz?: XYZDriver["setOptions"];
};

// [types.ts]

interface TransactionOptions {
  ttl?: number;
}

interface StorageDefinition {
  items: unknown;
  mounts: Record<string, { driver: keyof BuiltinDrivers }>; // <-- new
}

interface Storage<T> {
  getItem: (key: string, opts?: TransactionOptions & GetOptions) => any;
  setItem: (
    key: string,
    value: any,
    opts?: TransactionOptions & SetOptions
  ) => void;
}
```

- 1. Extracting get/set options as individual interfaces from each driver and auto-generating is a must-have that we are missing today (note that it is not tied to internal storage definition, but a public interface) 
2. `TypedStorage` and inferring the mount from the key could be done in the next steps to replace `Get/SetOptions` with an inferred subset from only the known vendor key. But please note it is really tricky, as currently we are maintaining a backward compatible `StorageItem` interface until the next major, and the benefit would be lower and requires users to explicitly type mount points as vendor drivers.

PR #662, opened 2025-08-24
feat(drivers): enhance drivers with typed options for `get`, `set`, `remove`, and `list` operations via generation

### LINKED ISSUE #660: Support strongly typed options in storage methods

schplitt:
Resolves #660

This is the discussed solution for typed options and would replace PR #661 and should be a very good fit with #610.

A Driver can specify a `<driverName>Driver` interface with set, get, remove and list options typed.
This can be used to Infer the complete operation options via the driver name so that we have strongly typed internal driver development.
This Driver is added to the generated _driver.ts and exposes all set, get, remove and list options from all drivers collectively.
Additionally there are now common operation options which all drivers can implement if they support them (like `ttl` for `set`).
Please check that get, set, remove and list make sense as common options or if we need to add/remove some.

~~I added a utility type to mimic the `safeName` from the `gen-drivers.ts` file. Drivers need to also respect that the options could be passed in via the safe name. I am unsure if this is the best solution. Let me know if you have any suggestions.~~

Please also see the `Storage` type to see how the new types are used.

This should not be a breaking changes as we still allow any other untyped properties on the root object like before.

If you have any questions, please feel free to ask right away.


schplitt:
Docs are currently missing. Will write them once the changes are final 😄

src/_drivers.ts
```diff
@@ -140,3 +142,28 @@ export const builtinDrivers = {
...
 } as const;
+
+export type BuiltinDrivers = {
+  "deno-kv": DenoKVDriver;
+  "denoKV": DenoKVDriver;
+}
+
+export type DriverGetOptions = {
+  "deno-kv"?: DenoKVDriver extends { getOptions: infer TGet } ? unknown extends TGet ? {} : TGet : {}
+  "denoKV"?: DenoKVDriver extends { getOptions: infer TGet } ? unknown extends TGet ? {} : TGet : {}
```
pi0, line 153: we should only keep one casing `denoKV: { ... }` to reduce runtime complexity of handling them also.

src/drivers/deno-kv.ts
```diff
@@ -12,15 +13,17 @@ export interface DenoKvOptions {
...
+
+export interface DenoKvDriver {
+  setOptions: {};
+  getOptions: {};
+  removeOptions: {};
 }
 
 const DRIVER_NAME = "deno-kv";
 
+type MethodTypes = InferOperationOptions<DenoKvDriver, typeof DRIVER_NAME>;
```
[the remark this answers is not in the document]
pi0: The input type to drivers is "agreegated", it could look like `{ ttl: 10, denoKV: { foo: "bar" }, cloudflareKV: { bar: "baz" } }`, we should type it from the generated namespace instead of inference.  (sorry if also i am not clear enough, i can happily push some refactors later if you want me)

docs/1.guide/5.developing-drivers.md
```diff
@@ -0,0 +1,120 @@
...
+Take a look at the [deno driver](https://github.com/unjs/unstorage/blob/main/src/drivers/deno-kv.ts) for a complete implementation example with typed options.
+
+## 5. Generating and Using Types Globally
+
+After defining your driver and its types:
+
+**Run the code generation script:**
+
+```sh
+pnpm gen-drivers
```
pi0, line 95: This looks like a mixed guide of making internal drivers (which should be in contribtion guide not here) and external drivers (which is unrelavant at least to this PR)

src/_drivers.ts
```diff
@@ -140,3 +142,24 @@ export const builtinDrivers = {
...
   "vercelRuntimeCache": "unstorage/drivers/vercel-runtime-cache",
 } as const;
+
+export type BuiltinDrivers = {
+  "deno-kv": DenoKVDriver;
+  "denoKV": DenoKVDriver;
+}
+
+export type DriverGetOptions = {
+  "deno-kv"?: DenoKVDriver extends { getOptions: infer TGet } ? unknown extends TGet ? {} : TGet : {}
```
pi0, line 152: This is not easily doable in runtime to pass `{ 'deno-kv': { ... opts } }`, we should instead use `{ denoKV: {} }` to be easier with JS objects.

src/types.ts
```diff
@@ -123,58 +160,52 @@ export interface Storage<T extends StorageValue = StorageValue> {
...
-    opts?: TransactionOptions
+    opts?: SetOptions
   ) => Promise<void>;
-  removeMeta: (key: string, opts?: TransactionOptions) => Promise<void>;
+  removeMeta: (key: string, opts?: RemoveOptions) => Promise<void>;
   // Keys
   getKeys: (base?: string, opts?: GetKeysOptions) => Promise<string[]>;
   // Utils
+
+  // TODO: what options should it have?
```
pi0, line 208: We probably need a placeholder mapping for `clear` too :D

src/_drivers.ts
pi0: PS: We could simplify the interface to the drivers to actually accept a custom option. We can start by denoKV as an example, but then we need to allow actually `tOpts.denoKV.*`.

Like discussed, inference of name could be in followup PR. This is mainly to document a "bag of possibe options" for get/set/etc

docs/1.guide/5.developing-drivers.md
schplitt: I created the CONTRIBUTION.md file instead.

I took inspiration from the guide of nitro.

PR #666, opened 2025-08-28
fix(prefixStorage): strip prefix from `keys()` 

RihanArfan:
This applies the same prefix stripping done to `getItems()` to it's alias `keys()`
https://github.com/unjs/unstorage/blob/da88311064cfb4bf41c4114b015c5bcb50906c8c/src/storage.ts#L474-L475

`prefixStorage()` is used with Nitro Storage


pi0:
Would be nice if we could add couple of tests covering this.

Issue #699, opened 2025-10-31
[v1] TypeScript types issue: Masquerading as CJS


kricsleo:
### Environment

`v1.17.1` (`v1` branch)

### Reproduction

Check for TypeScript type issues using [arethetypeswrong](https://github.com/arethetypeswrong/arethetypeswrong.github.io).

Run `npm run build && attw --pack .`

<img width="1484" height="309" alt="Image" src="https://github.com/user-attachments/assets/281238ac-bf47-42c0-ac67-cdb821e89a40" />

### Describe the bug

Imports with `"moduleResolution": "node16"` in tsconfig can cause issues such as https://github.com/unjs/unstorage/issues/698.

### Logs

```sh

```

Issue #698, opened 2025-10-30
Incorrect driver types


florian-lefebvre:
### Environment

node v22.19.0
unstorage v1.17.0 (1.17.1 in lockfile)

### Reproduction

Apologies for not being able to give a minimal repro. I'm using the fs lite driver for the Astro Fonts API and the types behave weirdly. To reproduce:
- Clone the astro repo https://github.com/withastro/astro
- Install deps (`pnpm i --frozen-lockfile`)
- Go to https://github.com/withastro/astro/blob/main/packages/astro/src/assets/fonts/implementations/storage.ts
-  Update the file contents to the following:

```ts
import { fileURLToPath } from 'node:url';
import { createStorage, type Storage } from 'unstorage';
import fsLiteDriver from 'unstorage/drivers/fs-lite';

export function createFsStorage({ base }: { base: URL }): Storage {
	return createStorage({
		driver: fsLiteDriver({
			base: fileURLToPath(base),
		}),
	});
}
```

### Describe the bug

I get the following type error: `This expression is not callable.
  Type 'typeof import("/home/florian/Documents/github/withastro/astro/node_modules/.pnpm/unstorage@1.17.1_@azure+identity@4.13.0_@netlify+blobs@10.2.1_@vercel+functions@2.2.13/node_modules/unstorage/drivers/fs-lite")' has no call signatures.ts(2349)`.

Updating the call works but this doesn't match the docs at https://unstorage.unjs.io/drivers/fs#nodejs-filesystem-lite:

```diff
import { fileURLToPath } from 'node:url';
import { createStorage, type Storage } from 'unstorage';
import fsLiteDriver from 'unstorage/drivers/fs-lite';

export function createFsStorage({ base }: { base: URL }): Storage {
	return createStorage({
-		driver: fsLiteDriver({
+		driver: fsLiteDriver.default({
			base: fileURLToPath(base),
		}),
	});
}
```

### Logs

```sh

```

faradaytrs:
in my case i typescript cant find types for cloudflare-kv-http at all.

pi0:
@faradaytrs can you share exact version and your taconfig?

PR #700, opened 2025-10-31
build: fix masquerading as CJS issue

kricsleo:
resolves https://github.com/unjs/unstorage/issues/699
~~resolves https://github.com/unjs/unstorage/issues/698~~

Before:
Image

After:
Image


pi0 (approved):
Thanks ❤️

PR #707, opened 2025-11-06
feat: experimental tracing channels support

logaretm:
Adds [Node.js tracing channels](https://nodejs.org/api/diagnostics_channel.html#tracingchanneltracepromisefn-context-thisarg-args) for all storage operations, enabling observability and performance monitoring.

### Channel Naming Convention

All channels use the `unstorage.${operation}` convention which when coupled with tracing channels produce the following diagnostic channels:

- `tracing:unstorage.${operation}:start`
- `tracing:unstorage.${operation}:end`
- `tracing:unstorage.${operation}:asyncStart`
- `tracing:unstorage.${operation}:asyncEnd`
- `tracing:unstorage.${operation}:error`

Which [matches nicely](https://fastify.dev/docs/latest/Guides/Migration-Guide-V5/#diagnostic-channel-support) with `fastify` but there isn't a lot of convention for them other than the [Node.js recommendation](https://arc.net/l/quote/rvrguopg).

### Traced operations

- Item operations: `hasItem`, `getItem`, `setItem`, `setItems`, `removeItem`
- Raw operations: `getItemRaw`, `setItemRaw`
- Meta operations: `getMeta` (uses `getItem` channel with `meta: true`)
- Batch operations: `getItems`, `setItems`
- Utility operations: `getKeys`, `clear`

### Tracing context includes

- `keys[]` - keys involved in the operation
- `base` - mount point handling the operation
- `driver` - driver name and options (single-key operations only)
- `meta` - flag for metadata operations

We can probably add more, in the Nuxt unstorage instrumentation we only use those.

### Example

```typescript
import { tracingChannel } from 'node:diagnostics_channel';

const channel = tracingChannel('unstorage.getItem');

channel.subscribe({
  start: (data) => {
    // start timing, or spans or whatever...
  },
  asyncEnd: (data) => {
    console.log(`${data.driver?.name} fetched ${data.keys[0]} in ${duration}ms`);
  },
  error: (data) => {
    console.error(`Error in ${data.driver?.name}:`, data.error);
  }
});
```

All operations emit standard tracing events: `start`, `end`, `asyncStart`, `asyncEnd`, and `error`.

---

**⚠️ Current Blocker:** unstorage can be used in browser, so we need to shim it in this case, or perhaps provide an alternative.


logaretm:
@pi0 I refactored the logic here to use `process.getBuiltIn...`, separated the tracing into its own subpath export, and reduced the logic size by generalizing a trace wrapper.

PR #733, opened 2026-01-17
chore(deps): remove anymatch

### LINKED ISSUE #732: Replace `anymatch` dependency with direct `picomatch` usage
### Describe the feature

- `anymatch` is no longer maintained
- `anymatch` depends on an old version of `picomatch`

### Additional information

gameroman:
Closes #732

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,7 +42,6 @@
     "unstorage": "pnpm jiti src/cli"
   },
   "dependencies": {
-    "anymatch": "^3.1.3",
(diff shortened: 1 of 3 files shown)


pi0:
Thanks for PR ❤️ We have migrated to node matched glob for next major (https://github.com/unjs/unstorage/commit/c7f915fc8916209ce74f510bdf68f3f2a0562e4d), however doing this backward can cause regressions if users don't use latest 20.x/22.x which is risky to backport.

PR #760, opened 2026-03-20
fix(s3): respect prefix parameter and add pagination for getKeys()

guoyangzhen:
## Problem

Two bugs in the S3 driver:

### Bug 1: prefix parameter ignored
 accepts a prefix parameter but never uses it in the fetch URL. When  is called, it returns ALL keys in the bucket instead of only keys matching the  prefix.

The function even logs the intended URL () but the actual fetch always uses  without any query parameters.

### Bug 2: no pagination
S3 ListObjectsV2 API returns max 1000 objects per response. Buckets with >1000 objects only return the first 1000 keys, silently dropping the rest.

## Fix

1. **Prefix**: Build the ListObjectsV2 URL with  query parameters
2. **Pagination**: Loop using  from the XML response until  is false
3. **Renamed**  to  to return both keys and pagination metadata

This fix also benefits  which internally calls .

## Summary by CodeRabbit

* **Bug Fixes**
  * S3 object listing now fully supports pagination and correctly retrieves all objects across multiple pages.
  * XML responses and pagination tokens are handled more robustly to avoid truncation and decoding errors.

* **New Features**
  * Bulk delete operations are more reliable and efficient: large deletes are chunked and individual deletes use bounded concurrency to reduce failures and resource spikes.


guoyangzhen:
All 3 CodeRabbit suggestions addressed:

1. ✅ **Fail-fast on truncated missing token** — Added validation in `parseListResponse` (commit e29b17f)
2. ✅ **XML entity decoding** — Added `decodeXmlText()` for keys and continuation tokens (commit e29b17f)
3. ✅ **Batch DeleteObjects for 1000-key limit** — Chunked bulk deletes into 1000-key batches, bounded per-object fallback to 10 concurrent deletes (commit db8e897)

The pagination change means `listObjects` can now return arbitrarily large key sets. S3's `DeleteObjects` API has a hard limit of 1000 keys per request, so bulk deletes must be chunked. For the `bulkDelete: false` fallback path, unbounded `Promise.all` is replaced with sequential batches of 10 to avoid overwhelming the provider.

guoyangzhen:
Hi, just checking in on this PR. Happy to make any changes if needed. Thanks!

PR #767, opened 2026-03-31
feat(fs,fs-lite): add dataSuffix option to prevent file/directory key collisions

claygeo:
## Summary

When both `foo` and `foo:bar` exist as storage keys, the `fs` and `fs-lite` drivers map them to `<base>/foo` (file) and `<base>/foo/bar` (inside directory `foo/`). This creates an ENOTDIR collision because `foo` cannot be both a file and a directory on disk.

This PR adds a `dataSuffix` option to both drivers. When set (e.g. `dataSuffix: ".data"`):

- Key `foo` is stored as `foo.data` on disk
- Key `foo:bar` is stored as `foo/bar.data` on disk
- No collision -- `foo.data` is a file, `foo/` is a directory

The suffix is transparently stripped from keys returned by `getKeys()` and watch callbacks. The option defaults to `undefined` for full backward compatibility.

## Context

This addresses the root cause of file/directory collisions reported in:
- nitrojs/nitro#4142
- nitrojs/nitro#4156

Per [pi0's feedback](https://github.com/nitrojs/nitro/pull/4156#issuecomment-4160482020), the fix belongs in the unstorage fs driver rather than in Nitro's cache key sanitization.

## Changes

- **`src/drivers/fs.ts`** -- Added `dataSuffix` option, split `r()` into `r()` (directory resolution) and `rFile()` (file resolution with suffix), updated all file operations to use `rFile()`, strip suffix in `getKeys()` and watch callbacks
- **`src/drivers/fs-lite.ts`** -- Same changes as `fs.ts`
- **`test/drivers/fs.test.ts`** -- Added test verifying `foo` + `foo:bar` coexistence with `dataSuffix`, and basic CRUD with suffix
- **`test/drivers/fs-lite.test.ts`** -- Same tests as `fs.test.ts`

## Usage

```ts
import fsDriver from "unstorage/drivers/fs";

  driver: fsDriver({ base: "./cache", dataSuffix: ".data" }),
});

// These now coexist without ENOTDIR errors:
await storage.setItem("foo", "value1");
await storage.setItem("foo:bar", "value2");
```

## Summary by CodeRabbit

* **New Features**
  * Added optional `dataSuffix` to filesystem storage drivers (`fs`/`fs-lite`) to prevent on-disk collisions (e.g., `foo` vs `foo:bar`) via suffixed leaf files.
  * `getKeys()` and change notifications now return logical keys with the suffix stripped.
* **Bug Fixes**
  * Added validation for invalid `dataSuffix` values and ensured the empty/root key can’t escape the storage base.
* **Tests**
  * Added coverage for coexistence, suffix stripping, ignoring non-matching files, validation errors, and watcher/update behavior.
* **Documentation**
  * Documented `dataSuffix`, collision scenarios, constraints, and related behaviors (including `clear()`).

### DIFF
--- a/docs/2.drivers/fs.md
+++ b/docs/2.drivers/fs.md
@@ -28,6 +28,7 @@ const storage = createStorage({
 - `base`: Base directory to isolate operations on this directory
 - `ignore`: Ignore patterns for watch <!-- and key listing -->
 - `watchOptions`: Additional [chokidar](https://github.com/paulmillr/chokidar) options.
+- `dataSuffix`: Suffix appended to on-disk filenames to prevent key/namespace collisions (see [below](#preventing-keynamespace-collisions)).
 
 ## Node.js Filesystem (Lite)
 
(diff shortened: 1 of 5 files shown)


Norbiros:
@pi0 (sorry for the direct ping), but would you mind taking a look at this PR? The issue it resolves directly impacts Nitro and, as a result, real Nuxt applications. Thanks in advance!

claygeo:
Polished this and re-confirmed it's still needed: the released and current `main` `fs`/`fs-lite` drivers still hit the `foo` + `foo:bar` file/directory collision (ENOTDIR on POSIX, ENOENT on Windows), verified at runtime.

This push adds:
- **Validation** of `dataSuffix` at construction — rejects empty, `.`/`..`, and any value containing `/`, `\` or `:` (these would create nesting, escape the base directory, or break the `getKeys`/watch round-trip).
- A **base-escape fix**: an empty/root key resolves to `base` itself, so `base + dataSuffix` wrote a sibling file *outside* the configured base directory (e.g. `<base>.data`). It now stays inside base.
- **Docs** for both drivers in `docs/drivers/fs`, including the on-disk migration caveat (enabling the option hides pre-existing un-suffixed files).
- **Tests**: validation, strip-exactly-once (a key already ending in the suffix), `getKeys` ignoring non-suffixed files, watcher suffix-stripping, and the root-key no-escape regression.

@pi0 this is the unstorage `fs`-driver fix you pointed to from nitrojs/nitro#4156 — ready for review when you have a moment.

claygeo:
Friendly nudge on this one, and a small update. I re-confirmed the file/directory key collision still reproduces on the latest `main` (keys `foo` and `foo:bar` can't coexist: `<base>/foo` would need to be a file and a directory), so this is still relevant.

Since the original PR I've hardened the `dataSuffix` implementation:

- **Validation** of `dataSuffix` at construction: rejects empty, `.`, `..`, and any value containing `/`, `\`, or `:` (these would create nesting, escape the base dir, or break the `getKeys`/watch round-trip).
- **Fixed a base-directory escape**: a root/empty key resolves to `base` itself, so `base + dataSuffix` wrote a sibling file *outside* the configured base (e.g. `<base>.data`). It now stays inside `base`.
- **Docs** for both `fs` and `fs-lite`, including the on-disk migration caveat and `clear()` behavior.
- **Tests**: collision coexistence, suffix validation, strip-exactly-once round-trip, watcher suffix stripping, and a root-key no-escape regression. All `fs`/`fs-lite` tests pass locally.

Still mergeable with no conflicts, default `undefined` keeps it fully backward compatible.

@pi0 you pointed nitrojs/nitro#4156 here as the right home for this fix. Would you (or another maintainer) be able to take a look when you get a chance? Happy to adjust naming or scope. Thanks!

PR #785, opened 2026-06-12
fix(utils): prefix watch callback strips prefix and filters external events

LeSingh1:
When you call `watch()` on a `prefixStorage` instance, two things go wrong today:

1. The callback receives the full key including the namespace prefix (e.g. `"ns:foo"` instead of `"foo"`), which is inconsistent with every other method on the prefixed storage.
2. Events for keys outside the prefix are delivered to the callback even though the caller only asked to watch a subset of the storage.

The root cause is that `prefixStorage` spreads the underlying storage object and then overrides individual methods, but `watch` was never overridden. Because of this, `nsStorage.watch` is just `storage.watch` -- the full-store watcher with no filtering or key transformation.

The fix adds a `watch` override on the prefixed storage that wraps the caller's callback. The wrapper checks whether each incoming key starts with the base prefix, drops the event if it does not, and strips the prefix from the key before invoking the callback. This makes `watch` consistent with how `getKeys`, `getItem`, and the other prefixed methods already behave.

A regression test covers both the key-stripping and the event-filtering behaviour.

## Summary by CodeRabbit

* **New Features**
  * Prefixed storage namespaces now support reactive change watching.
  * Watch callbacks receive only changes within the selected namespace, with namespace prefixes removed from keys for simpler integration.
  * Events from outside the namespace are excluded, helping applications respond only to relevant storage updates.


pi0x:
Thanks for the report and the fix @LeSingh1! 🙏

The diagnosis here was spot on, but this was overtaken by #793 which landed the same fix in `main` (commit e6be613) — `prefixStorage` now overrides `watch` to filter by base and strip the prefix, with test coverage in `test/storage.test.ts`.

Closing as superseded. Sorry for the slow turnaround on this one, and thanks again for taking the time!

PR #791, opened 2026-06-26
docs(azure): fix Azure Table Storage package name and complete options list

### LINKED ISSUE #681: Azure Table Storage docs are incomplete, and refer to invalid pkg @azure/data-table
### Describe the change

The [Azure Table Storage Driver](https://unstorage.unjs.io/drivers/azure#azure-table-storage) docs are incomplete. Notably, `sasKey`, `connectionString` and `pageSize` are not documented.

In addition, the docs refer to the package `@azure/data-table` which should be `@azure/data-tables`.

### URLs

https://unstorage.unjs.io/drivers/azure#usage-4

### Additional information

- [ ] Would you be willing to help?

MFA-G:
Fixes #681

The Azure Table Storage driver docs had a few inaccuracies, all verified against `src/drivers/azure-storage-table.ts`:

### Wrong package name
The install instructions referenced `@azure/data-table`, but the published package — and the driver's own import (`from "@azure/data-tables"`) — use the plural `@azure/data-tables`. Following the old instructions installs a non-existent package. Fixed both the prose and the `:pm-install` block.

### Broken link
Removed the dangling `[data-tables]()` line, which rendered as an empty link.

### Incomplete options list
The Options list was truncated mid-entry (a dangling `-`) and only documented `accountName`, `tableName`, `partitionKey` and `accountKey`. Added the missing options to match `AzureStorageTableOptions`:
- `sasKey` — SAS key for `AzureSASCredential`
- `connectionString` — connection string (Node.js only)
- `pageSize` — entries per request, defaults to `1000` (also the max, enforced in the driver)

### Auth note fix
The `AzureSASCredential` authentication bullet referenced a `sasToken` option that doesn't exist; the actual option is `sasKey`.

Docs-only change. No code touched.

## Summary by CodeRabbit

* **Documentation**
  * Clarified Azure Table Storage driver usage and setup guidance.
  * Updated the package install name to the current Azure Tables package.
  * Expanded authentication details, including SAS credential and Node-only connection string support.
  * Improved options guidance with clearer precedence rules and added the new `pageSize` setting with its default and maximum.


MFA-G:
Good catch. Fixed in 4a1059d — the precedence is `accountKey` → `sasKey` → `connectionString` (via `else if` fallthrough in the driver), so I updated the descriptions to say `accountKey` takes precedence over `sasKey`, and both take precedence over `connectionString` (matching the wording used in the Blob Storage section).

pi0 (approved):

PR #810, opened 2026-08-21
fix(azure-cosmos): dispose the cosmos client on dispose

### LINKED ISSUE #809: azure-cosmos driver never implements dispose(), leaks the CosmosClient's background endpoint refresher
azure-cosmos.ts only keeps a `Promise<Container>` around (the `client` closure variable), the actual `CosmosClient` instance created inside `getCosmosClient()` is never stored anywhere accessible outside that closure, and the driver object returned by this factory has no `dispose()` method at all.

@azure/cosmos's CosmosClient starts a background endpoint-refresher timer on construction. Its own type doc comment says: "Clears background endpoint refresher. Use client.dispose() when destroying the CosmosClient within another process." Without a dispose() implementation here, calling storage.dispose() on a storage instance using this driver never clears that timer, it leaks indefinitely and can keep a Node process alive.

I checked the other drivers with external connections (db0.ts, deno-kv.ts, mongodb.ts, redis.ts) and they all implement dispose(). PR #792 ("fix(mongodb): close client connection on dispose", merged) says explicitly "The redis, deno-kv and db0 (#693) drivers already close their connection on dispose; this brings mongodb in line" - azure-cosmos wasn't included in that pass.

Expected: azure-cosmos should expose the underlying CosmosClient (or wrap it) and call `.dispose()` on it when the driver's own dispose() runs.

lazerg:
The azure-cosmos driver only kept the resolved `Container` promise around, so the `CosmosClient` behind it was unreachable once the factory returned. `CosmosClient` starts a background endpoint refresher on construction and its own docs say to call `client.dispose()` when you are done with it, so `storage.dispose()` left that timer running and could keep a Node process alive.

The driver now holds on to the client and disposes it in its own `dispose()`, the same way #792 did for mongodb. The container promise is cleared too, so the driver reconnects if it gets used again.

Fixes #809

## Summary by CodeRabbit

* **New Features**
  * Added a disposal method for the Azure Cosmos connection, allowing resources to be cleanly released and the driver to be reused.
* **Bug Fixes**
  * Improved connection lifecycle handling, including cleanup during pending initialization and repeated disposal.
* **Tests**
  * Added coverage for disposal before connection, after connection, repeated disposal, and subsequent client recreation.

### DIFF
--- a/src/drivers/azure-cosmos.ts
+++ b/src/drivers/azure-cosmos.ts
@@ -6,7 +6,7 @@ import {
   type DriverDependencies,
 } from "./utils/index.ts";
 import { type AzureIdentityOptions, createDefaultAzureCredential } from "./utils/azure.ts";
-import type { Container } from "@azure/cosmos";
+import type { Container, CosmosClient } from "@azure/cosmos";
 
 export interface AzureCosmosOptions extends AzureIdentityOptions {
   /**
@@ -64,6 +64,7 @@ export interface AzureCosmosItem {
 
 const driver: DriverFactory<AzureCosmosOptions, Promise<Container>> = (opts) => {
   let client: Promise<Container> | undefined;
+  let cosmosClient: CosmosClient | undefined;
   const getCosmosClient = () =>
     (client ??= (async () => {
       if (!opts.endpoint) {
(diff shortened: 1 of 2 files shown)


lazerg:
Closing this since it has been inactive for a while.

PR #811, opened 2026-08-21
fix(azure-cosmos): dispose the CosmosClient on driver dispose

### LINKED ISSUE #809: azure-cosmos driver never implements dispose(), leaks the CosmosClient's background endpoint refresher

MFA-G:
Closes #809

### Problem

`azure-cosmos.ts` only kept the resolved `Promise<Container>`. The `CosmosClient` created inside `getCosmosClient()` never escaped that closure, and the driver exposed no `dispose()` at all.

`@azure/cosmos` starts a background endpoint refresher on construction, and its own type doc says:

> Clears background endpoint refresher. Use `client.dispose()` when destroying the CosmosClient within another process.

So `storage.dispose()` on a Cosmos-backed storage left that timer running — it leaks and can keep a Node process alive. The other drivers with external connections (`db0`, `deno-kv`, `mongodb`, `redis`) all implement `dispose()`; azure-cosmos was missed in the pass that added it to mongodb (#792).

### Change

- Hold the `CosmosClient` in a driver-scoped variable alongside the container promise.
- Add `dispose()`, mirroring the mongodb driver: await the in-flight connection so a pending attempt can't resurrect state, clear the cached promise, then call `client.dispose()`.

Disposing before ever connecting, and disposing twice, are both no-ops; using the driver again after dispose lazily creates a fresh client, same as mongodb.

### Validation

- `vitest run test/drivers/azure-cosmos.test.ts` → 1 passed (the pre-existing live-service suite stays `describe.skip`).
- The new test fails on `main` with `default(...).dispose is not a function`, so it pins the regression.
- The test drives the driver through a minimal `lib` stub, so it needs no Cosmos account and runs in CI.
- `oxlint` and `oxfmt --check` clean on both touched files.
- `tsc --noEmit --skipLibCheck` reports no new errors (the existing `unstorage/drivers/*` self-import errors are present on `main` too, they need a build first).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Azure Cosmos connection cleanup and resource disposal.
  * Prevented stale connections from being reused after disposal.
  * Ensured disposal is safe before a connection is established.
  * Added reliable cleanup when disposal occurs during connection initialization.
  * Enabled fresh connections to be created when the driver is reused.

### DIFF
--- a/src/drivers/azure-cosmos.ts
+++ b/src/drivers/azure-cosmos.ts
@@ -6,7 +6,7 @@ import {
   type DriverDependencies,
 } from "./utils/index.ts";
 import { type AzureIdentityOptions, createDefaultAzureCredential } from "./utils/azure.ts";
-import type { Container } from "@azure/cosmos";
+import type { Container, CosmosClient as CosmosClientType } from "@azure/cosmos";
 
 export interface AzureCosmosOptions extends AzureIdentityOptions {
   /**
@@ -64,6 +64,7 @@ export interface AzureCosmosItem {
 
 const driver: DriverFactory<AzureCosmosOptions, Promise<Container>> = (opts) => {
   let client: Promise<Container> | undefined;
+  let cosmosClient: CosmosClientType | undefined;
   const getCosmosClient = () =>
     (client ??= (async () => {
       if (!opts.endpoint) {
@@ -75,7 +76,7 @@ const driver: DriverFactory<AzureCosmosOptions, Promise<Container>> = (opts) => 
         opts.lib,
         () => import("@azure/cosmos"),
       );
-      const cosmosClient = opts.accountKey
+      cosmosClient = opts.accountKey
         ? new CosmosClient({ endpoint: opts.endpoint, key: opts.accountKey })
         : new CosmosClient({
             endpoint: opts.endpoint,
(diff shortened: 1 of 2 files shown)


MFA-G:
Thanks @coderabbitai — valid, added in ef7be75.

The existing tests only disposed *after* `getCosmosClient()` had resolved, so the `await client.catch(() => {})` in `dispose()` was never actually exercised — the in-flight path was the one thing the PR claims to handle and the one thing that was untested.

New test defers the `lib` import so initialization is genuinely pending when `dispose()` is called:

```ts
const pending = cosmosDriver.setItem!("a", "test_data", {});
expect(clients.length).toBe(0);
const disposed = cosmosDriver.dispose!();

releaseLib();
await pending;
await disposed;

expect(clients.length).toBe(1);
expect(clients[0]!.disposeCalls).toBe(1);
```

Verified it is a real regression guard: dropping `await client.catch(() => {})` from `dispose()` makes it fail with `disposeCalls` `1 → 0` (the client is constructed after `dispose()` returns and leaks), and it passes with the fix.

`vitest run test/drivers/azure-cosmos.test.ts` → 2 passed, no type errors; prettier clean.

PR #817, opened 2026-09-02
fix(storage): emit watch events for setItems

mahirhir:
### What goes wrong

`setItem`, `setItemRaw` and `removeItem` each end with

```ts
if (!driver.watch) {
  onChange("update", key);
}
```

so a `storage.watch()` callback sees the write whenever the mount's driver has no native watch. `setItems` has no such call, so **batch writes are invisible to watchers**.

That means anything built on `watch` — cache invalidation, HMR, mirrored state — silently goes stale on exactly the writes that touch the most keys, and it only shows up on drivers without a native watch (memory, and most others), which is the common case.

### The change

Emit one `update` per item after the batch lands, under the same `!driver.watch` condition the single-key paths use. Both branches are covered: the native `driver.setItems` path and the `setItem` fallback. Restructured the early `return`s into `if / else if / else` so a single event loop can follow either write path — no behavior change other than the events.

`removeItems` does not exist on `Storage`, so there is no matching gap on the remove side.

### Verification

Two tests in `test/storage.test.ts`, one per branch (memory has no `setItems`, so the second uses a small inline driver that implements it).

Negative control — the tests against unpatched `src/storage.ts`:

```
 × storage > watch: setItems
 × storage > watch: setItems with a driver implementing setItems
AssertionError: expected "vi.fn()" to be called with arguments: [ 'update', 'mnt:etc:conf' ]
Number of calls: 0
```

Zero calls on both paths, which is the bug. With the fix, `vitest run test/storage.test.ts` is 18/18 with no type errors, and `oxlint` / `oxfmt --check` are clean on both files.

Full-suite caveat: `test/drivers/netlify-blobs.test.ts` fails 6 tests here, but it fails identically on a pristine checkout (I stashed and re-ran to confirm), so it is unrelated to this change.

## Summary by CodeRabbit

- **Bug Fixes**
  - Multi-item storage updates now continue processing remaining items when one write fails.
  - Successful item updates still trigger notifications even when another item fails.
  - Write failures are still reported after all pending item updates finish.

- **Tests**
  - Added coverage for partial failures and update notifications during multi-item storage operations.
  - Expanded validation for native batch-update behavior.

### DIFF
--- a/src/storage.ts
+++ b/src/storage.ts
@@ -221,7 +221,7 @@ export function createStorage<T extends StorageValue>(
     async setItems(items, commonOptions) {
       await runBatch(items, commonOptions, async (batch) => {
         if (batch.driver.setItems) {
-          return asyncCall(
+          await asyncCall(
             batch.driver.setItems,
             batch.items.map((item) => ({
               key: item.relativeKey,
@@ -230,20 +230,25 @@ export function createStorage<T extends StorageValue>(
             })),
             commonOptions,
           );
+        } else if (batch.driver.setItem) {
+          await Promise.all(
+            batch.items.map((item) => {
+              return asyncCall(
+                batch.driver.setItem!,
+                item.relativeKey,
+                stringify(item.value),
+                item.options,
+              );
+            }),
+          );
+        } else {
+          return; // Readonly
         }
-        if (!batch.driver.setItem) {
-          return;
+        if (!batch.driver.watch) {
+          for (const item of batch.items) {
+            onChange("update", item.key);
+          }
         }
-        await Promise.all(
-          batch.items.map((item) => {
-            return asyncCall(
-              batch.driver.setItem!,
-              item.relativeKey,
-              stringify(item.value),
-              item.options,
-            );
-          }),
-        );
       });
     },
     async setItemRaw(key, value, opts = {}) {
--- a/src/storage.ts
+++ b/src/storage.ts
@@ -231,7 +231,9 @@ export function createStorage<T extends StorageValue>(
             commonOptions,
           );
         } else if (batch.driver.setItem) {
-          await Promise.all(
+          // A rejected sibling write must not suppress the events for the
+          // writes that did land: those values are in the driver either way.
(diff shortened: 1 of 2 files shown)


mahirhir:
Both review findings were right and are fixed in `1c3ee59`. The second one was a defect in my test rather than in the code, and it is the more interesting of the two.

**1. Events for writes that landed.** In the fallback path `Promise.all` rejects on the first failure, so the emit loop after it never runs, even for items the driver had already stored. Now the writes are settled, events are emitted for the fulfilled ones, and the failure is rethrown after.

**2. The test could not fail for its own reason.** The fixture in `watch: setItems with a driver implementing setItems` implemented both `setItem` and `setItems`, so a regression that routed the batch through the per-item path would write the same value and emit the same event, and the test would still pass. Its `setItem` now throws, and the test asserts `setItems` actually ran and stored the value.

I checked that rather than assuming it. Disabling the native branch (`if (false && batch.driver.setItems)`) against each version of the test:

```
                                      old test      new test
watch: setItems with a driver...      PASSED        FAILED
```

So the finding was exactly right: under the regression it was meant to catch, that test was green. One unrelated pre-existing test (`setItems doeesn't upload twice`) did catch it, so the repo was not blind, but the watch test specifically was not contributing.

The new sibling-failure test gets the same treatment. Reverting `Promise.allSettled` back to `Promise.all`:

```
AssertionError: expected "vi.fn()" to be called with arguments: [ 'update', 'mnt:data:good' ]
Number of calls: 0
```

Worth noting for anyone reading this later: my first attempt at that control silently did not apply the mutation, and the suite came back green. I read that as "the test does not catch it" for a minute before checking whether the file had actually changed. A mutation control needs to assert that it mutated something.

```
Tests  19 passed (19)
tsc --noEmit: no new errors (the 35 remaining are pre-existing
              self-referential `unstorage/*` imports that need a build)
```

The `Vercel` check is red because the deployment needs a team member to authorize it for an outside fork, not because of anything in the diff. `tests`, `codecov/patch`, `codecov/project` and `autofix` are all green.

